PR #7, opened 2020-08-10
Improvement/dashboard fixes

ahellander:
Improvement to the Dashboard. Added an overview table for model validations embedded in the Controller app. Resolves FEDN-5, FEDN-6, FEDN-11.


components/discovery/combiner/templates/index.html
```diff
@@ -23,7 +23,11 @@ <h2>Alliance information</h2>
         <div class="" style="min-height:200px;">
             <h2>Model performance</h2>
             <p>Track progress of the federated model.</p>
-              <a class="btn btn-primary" href="http://localhost:5111/box" role="button">Validations</a>
+
+            <div class="embed-responsive embed-responsive-21by9">
+                <iframe class="embed-responsive-item" src="http://localhost:5111/table" allowfullscreen></iframe>   
+            </div>
+            <a class="btn btn-primary" href="http://localhost:5111/box" role="button">Validations</a>
```
morganekmefjord, line 30: This will only work intermittently in our test env. We need to refactor this out to a configurable host.

components/discovery/seed.json
```diff
@@ -45,9 +45,12 @@
...
   "fields": {
     "combiner": 8,
+    "max_clients": 5,
     "timeout": 180,
-    "rounds": 5,
-    "clients_required": 2,
+    "rounds": 1,
+    "clients_required": 1,
+    "clients_requested": 5,
+    "algorithm": "fedavg",
```
morganekmefjord, line 53: Great if we can set this as you have introduced on a combiner level. Will the reducer be responsible for this instruction?

ahellander: Yes, this is what I am thinking. Either the Reducer or some other entity.

components/discovery/combiner/templates/index.html
ahellander: Yep, there is an issue in Jira. We need to decide if we keep the Dashboard app separate as now, or if we integrate in the controller app or alliance_admin app.

PR #10, opened 2020-08-13
New fixes for Data Center example including combiner and reducer logic

sztoor:


datacetner_clients.ymal
```diff
@@ -0,0 +1,50 @@
+version: '3.7'
```
ahellander, line 1: The filename and ending has "error". Can you change the name to "datacenter_clients.yaml" ?

test/data_center/combiner/init_model.py
```diff
@@ -0,0 +1,59 @@
+import keras
```
ahellander, line 1: In the last version in master, I changed the way we do the seeding for the mnist example. Instead of trying to write directly to Minio, we now produce the model file  and bundle it in the repo for convenience. Can you do the same for this example?  (see test/mnist-multi-combiner/seed/ ) and the updated README.md.

sztoor:
Now I have fixed the names and the new directory structure in data-center example. The example is working.

PR #22, opened 2020-09-07
Reducer control

ahellander:


fedn/fedn/clients/reducer/control.py
```diff
@@ -1,10 +1,62 @@
...
+import copy
+
+
+import fedn.common.net.grpc.fedn_pb2 as fedn
+import fedn.common.net.grpc.fedn_pb2_grpc as rpc
+import grpc
+
+from fedn.algo.fedavg import FEDAVGCombiner
 
 class ReducerControl:
```
morganekmefjord, line 11: This class got fat pretty quick, perhaps this should be moved to a nother layer of indirection to allow for alternate control strategies, hence moving to a ReducerControlStrategy

morganekmefjord (approved):

fedn/fedn/algo/fedavg.py
```diff
@@ -167,70 +153,112 @@ def run(self):
...
+                    # TODO - is this how we want to do it ?
+                    self.config = plan
+                    if plan['task'] == 'training':
+                        self.exec_training(plan)
+                        self.server.set_latest_model(self.model_id)
+                    elif plan['task'] == 'validation':
+                        self.exec_validation(plan)
+                    else:
+                        result = self.exec(plan)
+                        self.server.set_latest_model(self.model_id)
```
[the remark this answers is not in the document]
ahellander: Agree.

fedn/fedn/clients/reducer/control.py
ahellander: I think those imports are not actually used here, removing them. 

Yes, I think this should become abstract, hooking in implementations of "spread_model" and "resolve" (those names should be worked on as well). In the general case, these can / will also be completely decentralized protocols.

PR #27, opened 2020-09-15
Hierarchical FedAvg Reducer control strategy works.

ahellander:
This PR completes a first implementation of Hierarchical FedAvg. 

It also introduces a major change in fedavg.py - the Combiner does no longer write any of its intermediary or final models  to persistent storage. We will revisit the feature of checkpointing combiner rounds later.


morganekmefjord (approved):
I have tested it and it seems to work fine.

PR #42, opened 2020-09-30
permission problem in Ubuntu

saadiabadi:


ahellander:
Need some more details and some review of the language. Do we really need to set the permission to executable for user? Would it not be better to simple manually create the cert folder? @sztoor How did you solve this?

morganekmefjord (changes requested):
We setup the folders in the Dockerfile using a specific fednuser.
If there is a problem with that user and how it mounts to local dir it would be best to figure out a way to solve the fednuser permissions such that they are OK with local system share. 

https://github.com/scaleoutsystems/fedn/blob/master/components/client/Dockerfile

morganekmefjord:
@saadiabadi  PR #61 to fix the base image will likely resolve these issues you see. Can you test it out and close this if they are resolved?

PR #58, opened 2020-10-07
Fedn cleanup

morganekmefjord:
Stale links may cause a race condition when starting your docker-compose suite.
Pending a race condition has occured a manual cleaning of folders are required. This is tedious.

Removing the links as they are no longer required.


mnist-clients.yaml
```diff
@@ -17,9 +17,7 @@ services:
     working_dir: /app/
```
ahellander, line 17: I think we should remove this file and only use client.yaml (with --scale argument) for MNIST. Less to maintain.

combiner.yaml
```diff
@@ -46,50 +43,5 @@ services:
       dockerfile: components/client/Dockerfile
     working_dir: /app/client
     command: fedn run combiner -d ${FEDN_REDUCER_HOST} -p ${FEDN_REDUCER_PORT} -t ${FEDN_ALLIANCE_ADMIN_AUTH_TOKEN} -h combiner2 -i 12081 -n combiner2
-    volumes:
```
ahellander, line 49: For now, should we add a volume mount on host for the TempFileStorage to mitigate Combiner failure risk?

ahellander (approved):

morganekmefjord:
Yes I will make another update or two with findings I have made.

PR #59, opened 2020-10-08
Add new plot for time/round, add new dashboard to automatically monit…

aitmlouk:
- Add new plot for time/round;
- Add new dashboard to automatically monitor the evolving performance of  REST API endpoints


fedn/fedn/clients/reducer/restservice.py
```diff
@@ -191,7 +195,7 @@ def create_plot(feature):
             elif feature == 'timeline':
                 return create_timeline_plot()
             elif feature == 'ml':
```
ahellander, line 197: Maybe for clarity, change the 'ml' to something else.

fedn/fedn/clients/reducer/templates/index.html
```diff
@@ -126,7 +126,7 @@ <h3 class="text-muted">{{client}}</h3>
                     {% if not h_latest_model_id %}
                     <!-- enctype="multipart/form-data" important for file upload-->
                     <form id="seed_upload_form" method="post" action="/seed" enctype="multipart/form-data">
-                        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
```
ahellander, line 129: I think these serve a specific purpose, right @morganekmefjord . Should not be removed.

components/client/Dockerfile
```diff
@@ -11,3 +11,7 @@ RUN pip install -e /app/fedn
 RUN chown -R fednuser /app
 USER fednuser
 WORKDIR /app
+
+# Clone the dashboard files into the docker container and install package drom dev branch
+RUN git clone --branch development https://github.com/flask-dashboard/Flask-MonitoringDashboard.git
```
ahellander, line 16: I think we should remember to refactor here, so that we have separate images for clients, combiners, the server.

ahellander (changes requested):

fedn/fedn/clients/reducer/templates/index.html
aitmlouk: Yes, should be important for security enhancements, even in the endpoints part I ignore them just to test the new integrated dashboard, if it's useful we can go further and extend it with CSRF, otherwise we ignore the dashboard and continue with own dashboard.

components/client/Dockerfile
aitmlouk: This is also for the new dashboard, development branch to let CPU loads work (CPU not supported in pypi package).

ahellander (approved):

PR #60, opened 2020-10-09
Added Mapview integration.

dstoyanova:


fedn/fedn/clients/reducer/restservice.py
```diff
@@ -211,6 +211,56 @@ def _scalar_metrics(metrics):
...
+        @app.route('/map_view')
+        def map_view():
+            map = create_map()
+            try:
+                return render_template('index.html', show_map=True, map=map)
+            except Exception as e:
+                return str(e)
+
+        def create_map():
+            mapbox_access_token = "pk.eyJ1IjoiZHN0b3lhbm92YSIsImEiOiJja2cwbnBvYzUwMWFjMnhvNTY5azZ1NWs5In0.plMjsaU58g3jL9_50WSq9w"
```
ahellander, line 223: Is this OK to have in the code?

fedn/fedn/clients/reducer/templates/index.html
```diff
@@ -84,6 +84,15 @@
...
                                 Dashboard</a>
                         </li>
+                        <li class="nav-item">
+                            <a class="nav-link" href="/map_view" tabindex="-1" aria-disabled="true">
+                                <svg width="1em" height="1em" viewBox="0 0 16 16" class="bi bi-tv" fill="currentColor"
+                                     xmlns="http://www.w3.org/2000/svg">
+                                    <path fill-rule="evenodd"
+                                          d="M2.5 13.5A.5.5 0 0 1 3 13h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5zM13.991 3H2c-.325 0-.502.078-.602.145a.758.758 0 0 0-.254.302A1.46 1.46 0 0 0 1 4.01V10c0 .325.078.502.145.602.07.105.17.188.302.254a1.464 1.464 0 0 0 .538.143L2.01 11H14c.325 0 .502-.078.602-.145a.758.758 0 0 0 .254-.302 1.464 1.464 0 0 0 .143-.538L15 9.99V4c0-.325-.078-.502-.145-.602a.757.757 0 0 0-.302-.254A1.46 1.46 0 0 0 13.99 3zM14 2H2C0 2 0 4 0 4v6c0 2 2 2 2 2h12c2 0 2-2 2-2V4c0-2-2-2-2-2z"/>
+                                </svg>
+                                Map View</a>
```
ahellander, line 94: Suggest to change menu label to "Network"

ahellander (changes requested):

fedn/fedn/clients/reducer/restservice.py
dstoyanova: It's a token for the Mapbox API which is linked to my account there. It's not a security risk when it comes to FEDn but whoever has the token, can use the Mapbox API without creating their own account. I can add it as a Git secret if you give me access to the Settings page of this repository.

fedn/fedn/clients/reducer/templates/index.html
dstoyanova: Alright, I can rename that.

fedn/fedn/clients/reducer/restservice.py
morganekmefjord: Then it will only work for any build job we setup, is there an alternative with a trial that would work for anyone starting this on their own machine? An api to request the token or something similar?

morganekmefjord: If it is not possible to have the api working without a key I think we should look at alternatives like the openstreetmap/openlayers framework.

dstoyanova: I have now removed the Mapbox functionality and I have used `plotly.express` instead. No API access token is needed for it.

ahellander:
I think this map solution looks good and nice that there are no additional dependencies! Now we need to hook it up with the actual combiners.

ahellander:
This work has been merged into the combiner-network branch and will be included in that PR.

PR #63, opened 2020-10-09
Bug/fedn 98

ahellander:
Fixes FEDN-98

It correctly handles the case that a combiner fails to do a model update for any other reason than that is has died completely.


fedn/fedn/clients/reducer/control.py
```diff
@@ -164,36 +169,46 @@ def round(self, config):
...
             # TODO: Should we reset combiner state here?
+            print("REDUCER CONTROL: Round invalid!",flush=True)
             return None
 
         # 3. Reduce combiner models into a global model
         # TODO, check success
-        model = self.reduce(updated)
+        try:
+            model = self.reduce(updated)
+        except:
```
morganekmefjord, line 187: Can we narrow this down to Some specific exceptions?

morganekmefjord (approved):
The wise exception should be narrowed down if possible. Other wise go ahead.

morganekmefjord (approved):

PR #64, opened 2020-10-10
Introduce custom baseimage override

morganekmefjord:
To streamline better test projects In this suggested commit there are a new structure to allow for overriding of parameters.
```yaml
version: '3.7'
services:
  client:
    build:
      context: .
      dockerfile: components/client/Dockerfile
      args:
        baseimage: "tensorflow/tensorflow:latest"
```
Illustrated by an example
```bash
docker-compose -f docker-compose.yaml -f reducer.yaml -f combiner.yaml -f client.yaml **-f tensorflow.yaml** --build --scale client=5
```
where tensorflow.yaml in this example is overriding the BASEIMAGE variable to the Dockerimage and instructs which base image to initiate the client image from.

In this sense the tests we construct can have the same basic structure but depending on framework selected can _inherit_ base image from for example tensorflow to ensure the execution context is present.

### DIFF
--- a/components/client/Dockerfile
+++ b/components/client/Dockerfile
@@ -1,13 +1,9 @@
-FROM morganekmefjord/basesci:latest
-RUN adduser --disabled-password --gecos '' fednuser \
-    && adduser fednuser sudo \
-    && echo '%sudo ALL=(ALL:ALL) ALL' >> /etc/sudoers
-
+ARG baseimage="python:3.8"
+FROM ${baseimage}
 
 RUN mkdir -p /app/client
 RUN mkdir /app/certs
 COPY fedn /app/fedn
+RUN pip install numpy
 RUN pip install -e /app/fedn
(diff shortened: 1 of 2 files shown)


ahellander:
Looks good but you accidentally made the PR to master...

morganekmefjord:
> Looks good but you accidentally made the PR to master...

Changed base to develop..

morganekmefjord:
@ahellander But what do you think of this  as an alternative approach to handle underlaying framework dependencies?

sztoor:
If we want to support different versions of TensorFlow, I think it will be a good strategy. But it may introduce extra complexity in the framework.

ahellander:
Yes, I think this is a good idea. Also because we will need to link in a different environment depending on the helper class we are using (TF, Torch, Sklearn) also for the Combiners.

PR #66, opened 2020-10-12
Working test case of Data Center with compute as a package.

sztoor:
The data center test case is now working with the latest changes.


ahellander:
Tested and pushed a minor fix - only two clients were used since three of them had the same name.

ahellander (approved):

PR #70, opened 2020-10-13
Abstract helper

ahellander:
Refactored ML helpers, introduce abstract base class. To improve the process to extend the framework with custom helpers. 

There is also a lot of additions from Li related to work in progress on a torch-helper, but you can ignore them in the review for now, will be updated later. The key thing to look at @morganekmefjord is how you feel about the helper abstract class.


fedn/fedn/clients/reducer/control.py
```diff
@@ -3,11 +3,10 @@
 import tempfile
 import time
 
-from fedn.utils.helpers import KerasSequentialHelper
 from fedn.clients.reducer.interfaces import CombinerUnavailableError
 
 from .state import ReducerState
-
+from abc import ABC,abstractmethod
```
morganekmefjord, line 9: I think this can be removed as it is not used.

test/pytorch_mnist/data/clean
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+rm data* -rf
```
morganekmefjord, line 2: for this to work you need to change the position of arguments recursive/force '-rf' to $rm -rf data*

test/pytorch_mnist/data/prepare
```diff
@@ -0,0 +1,12 @@
...
+wget http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
+wget http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz
+
+gunzip train-images-idx3-ubyte.gz
+gunzip train-labels-idx1-ubyte.gz
+
+echo Image data downloaded
+python3 process_mnist.py
+
+rm train-* -rf
```
morganekmefjord, line 12: rm -rf train-*

morganekmefjord (changes requested):
Some minor changes for `rm -rf` arguments then go ahead and merge. Nice work!

PR #76, opened 2020-10-26
NLP use case -- sentiment analysis task (50k) with Glove pre-trained word embeddings

aitmlouk:
- NLP use case using IMDB dataset (preprocessed data, text processing pipeline)
- Customise alliance timeline
- Count the nbr of models for a large number of rounds (delete db and bucket from seed UI)


ahellander (commented):
Looking good, but we need to find a way for the clients to grab the datasets before merging. We cannot keep using GitLFS. I think it is an acceptable solution to provide download instructions in the readme.md file in /test/nlp_imdb_senti/README.md  and let the user download and stage the dataeset. This is how a "real" client would do it (except they would already have the data locally).

ahellander (approved):

PR #341, opened 2022-01-28
Sk 56

Wrede:
## Status

- [x] Ready
- [ ] Draft
- [ ] Hold

## Description
- Fixes issue that client can still get assigned though no compute package or seed model has been configured.
  "checkers" functions are introduced which check the compute package and seed model in REST /assign. 
- these functions are now also used when "redirecting" URLs to /setup and /setup_model when not configured.

- Introduces new a feature --remote for reducer, enabling only local compute package when --remote is used for client.
  I.e no UI for uploading compute package is needed in this mode/setting. However, the compute package must be "mounted" on each local client. 
- Various other fixes and features (CLI)  associated to the above. 
- Unittests for the newly implemented functions. 

Note: 
control.state in statestore is not used...?

## Types of changes

What types of changes does your code introduce to FEDn?

- [ ] Hotfix (fixing a critical bug in production)
- [x] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist

_If you're unsure about any of the items below, don't hesitate to ask. We're here to help! 
This is simply a reminder of what we are going to look for before merging your code._

- [x] This pull request is against **develop** branch (not applicable for hotfixes)
- [ ] I have included a link to the issue on GitHub or JIRA (if any)
- [ ] I have included migration files (if there are changes to the model classes)
- [ ] I have read the [CONTRIBUTING](https://github.com/scaleoutsystems/fedn/blob/master/CONTRIBUTING.md) doc
- [ ] I have included tests to complement my changes
- [ ] I have updated the related documentation (if necessary) 
- [ ] I have added a reviewer for this pull request
- [ ] I have added myself as an author for this pull request

## Further comments

Anything else you think we should know before merging your code!

### DIFF
--- a/fedn/cli/run_cmd.py
+++ b/fedn/cli/run_cmd.py
@@ -44,6 +44,15 @@ def set_token(init_config, token):
             #no token was set by flag, generate one
             return str(uuid.uuid4())
 
+def check_helper_config_file(config):
+    control = config['control']
+    try:
+        helper = control["helper"]
+    except KeyError:
+        print("--remote was set to False, but no helper was found in --init settings file.", flush=True)
+        exit(-1)
+    return helper
+
 @main.group('run')
 @click.pass_context
 def run_cmd(ctx):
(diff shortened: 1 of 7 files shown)


ahellander:
Also, lets think about the naming of this option. 

--remote=False/True is not intuitive, it is not apparent that this means that one uses a local package. Especially on the reducer side it becomes unintuitive.  

Maybe something like 
--local-package  ? 

Also, I think that one should not be allowed to use a local package with a reducer that has a package set.

fedn/cli/run_cmd.py
```diff
@@ -44,6 +44,15 @@ def set_token(init_config, token):
             #no token was set by flag, generate one
             return str(uuid.uuid4())
 
+def check_helper_config_file(config):
+    control = config['control']
+    try:
+        helper = control["helper"]
+    except KeyError:
+        print("--remote was set to False, but no helper was found in --init settings file.", flush=True)
```
ahellander, line 52: --remote changed, update message

ahellander (changes requested):

fedn/cli/run_cmd.py
Wrede: fixed

ahellander (approved):

PR #364, opened 2022-03-04
Feature/docker conditional req

mcapuccini:
## Status

- [X] Ready
- [ ] Draft
- [ ] Hold

## Description
The current Dockerfile installs Keras deps by default. This PR makes requirements installation conditional. In this way the compose file install deps only in the clients.


mcapuccini:
@ahellander the CI is passing now!

ahellander (approved):

PR #380, opened 2022-04-08
Gramine compatibility patch

mcapuccini:
## Description
This patch enables client compatibility with Gramine. The `preserve_times` flag set to false avoids calling the unimplemented `utime` system call in the LibOS. It doesn't seem like preserving times when copying the files from the package is crucial for FEDn, but please feel free to propose another solution.

### DIFF
--- a/fedn/fedn/common/control/package.py
+++ b/fedn/fedn/common/control/package.py
@@ -215,7 +215,7 @@ class PackageRuntime:
         from_path = os.path.join(os.getcwd(), 'client')
         
         from distutils.dir_util import copy_tree
-        copy_tree(from_path, run_path)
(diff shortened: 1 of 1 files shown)


fedn/fedn/common/control/package.py
```diff
@@ -215,7 +215,7 @@ def dispatcher(self, run_path):
         from_path = os.path.join(os.getcwd(), 'client')
         
         from distutils.dir_util import copy_tree
-        copy_tree(from_path, run_path)
+        copy_tree(from_path, run_path, preserve_times=False)
```
ahellander, line 218: Maybe add a comment / note about why that flag is set in case we need to come back to it later.

ahellander (approved):

PR #381, opened 2022-04-12
Make all links static in README

mcapuccini:
## Description
This PR makes all links in the README.rst statics. This is needed as we use the same RST files in both the documentation and the GitHub root README.


Wrede (changes requested):
Can we also fix the other links to "deployment", "architecture" etc?

mcapuccini:
@Wrede I think I got them right now! :)

PR #384, opened 2022-05-02
Image build refactory

mcapuccini:
## Description
This PR introduces the following improvements/changes on Docker image build:
- Use the root image with venv instead of components/base/Dockerfile
- Reduce code redundancy by having a single workflow
- Push only to GitHub registry
- Push for master, develp and release only (PR are just built to see if there is any problem with the Dockefile)
- Push images with examples dependency


Wrede (approved):
Looks good!

PR #388, opened 2022-05-18
ADD token authentication to UI

mcapuccini:
## Description
This PR adds token authentication to the UI. The first time the user needs to add `/?token=<token>` to URL. The token is then set to a cookie and checked for each view.


Wrede:
If we want to use this in stackn, we will need to be able to set the token in CLI or by a env var.

Wrede (approved):
Looks good, however see my previous comment.

PR #392, opened 2022-05-23
Add instructions to connect examples to distributed deployment

mcapuccini:
## Description
This PR adds instructions to connect examples to remote deployment. A minor modification as also been made to the examples' `fedn.yaml` in order to pass the data path. This will also bypass the docker logic such that mounting the socket won't be necessary.


examples/mnist-keras/README.md
```diff
@@ -52,4 +53,25 @@ bin/build.sh
...
+> **Note** make sure to replace `<reducer-fqdn>` and `<reducer-port>` with appropriate values.
+
+Now you are ready to start the client via Docker by running the following command.
+
+```bash
+docker run -d \
+  -v $PWD/client.yaml:/app/client.yaml \
+  -v $PWD/data:/var/data \
+  -e ENTRYPOINT_OPTS=--data_path=/var/data/mnist.npz \
+  --add-host=reducer:<reducer-ip> \
```
Wrede, line 73: reducer and combiner hostnames are specific to our docker-compose. If the user have deployed reducer and combiners separately without compose, these names might be different.

examples/mnist-keras/README.md
```diff
@@ -52,4 +53,25 @@ bin/build.sh
...
+  --add-host=combiner:<combiner-ip> \
+  ghcr.io/scaleoutsystems/fedn/fedn:develop-mnist-keras run client -in client.yaml
+```
+> **Note** make sure to replace `<reducer-ip>` and `<combiner-ip>` with appropriate values.
```
Wrede, line 77: Perhaps add here that --add-host is required if reducer and combiner names are not external dns allowing ingress.

Wrede (changes requested):

examples/mnist-keras/README.md
ahellander: We do not need to map the reducer since the fqdn is set in the config file.

examples/mnist-keras/README.md
mcapuccini: Good point... what could be a good way to point this out in the docs? What determine these, DNS?

ahellander (approved):

PR #396, opened 2022-06-07
Fix PyTorch example data mount path in compose file

mcapuccini:
## Description
Fix PyTorch example data mount path in override compose file.


Wrede:
integration tests fail expected?

mcapuccini:
@Wrede the CI is broken but I fix it in #397. Please review that and than I will update this PR.

Wrede (approved):

PR #400, opened 2022-06-09
Increase number of retries in CI test

mcapuccini:
## Description
Increase number of retries in CI test and some formatting.


mcapuccini:
depends on #400

ahellander (approved):

PR #405, opened 2022-06-15
Replace MD5 checksum with SHA256

mcapuccini:
## Description
Replace MD5 checksum with SHA256. Implements linked pentest recommendation.

### DIFF
--- a/fedn/fedn/clients/reducer/restservice.py
+++ b/fedn/fedn/clients/reducer/restservice.py
@@ -22,6 +22,7 @@ from werkzeug.utils import secure_filename
 from fedn.clients.reducer.interfaces import CombinerInterface
 from fedn.clients.reducer.plots import Plot
 from fedn.clients.reducer.state import ReducerState, ReducerStateToString
+from fedn.utils.checksum import sha
 
 UPLOAD_FOLDER = '/app/client/package/'
 ALLOWED_EXTENSIONS = {'gz', 'bz2', 'tar', 'zip', 'tgz'}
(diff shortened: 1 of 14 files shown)


fedn/fedn/clients/reducer/restservice.py
```diff
@@ -958,10 +958,10 @@ def checksum():
 
             file_path = os.path.join(UPLOAD_FOLDER, name)
             print("trying to get {}".format(file_path))
-            from fedn.utils.checksum import md5
+            from fedn.utils.checksum import sha
```
Wrede, line 961: I know there's a lot of this in fedn, but perhaps we can start to move imports to the beginning of the file?

Wrede (changes requested):

fedn/fedn/clients/reducer/restservice.py
Wrede: that way it will also be handled by isort

mcapuccini: fixed

fedn/fedn/common/control/package.py
```diff
@@ -164,11 +164,11 @@ def validate(self, expected_checksum):
         """
         self.expected_checksum = expected_checksum
 
-        from fedn.utils.checksum import md5
+        from fedn.utils.checksum import sha
```
Wrede, line 167: here as well

Wrede (changes requested):

fedn/fedn/common/control/package.py
mcapuccini: done

Wrede (approved):

PR #415, opened 2022-07-05
Validate user-specified name for components (XSS vulnerability)

mcapuccini:
## Description
Component names validations to avoid XSS vulnerability.


Wrede (approved):
great!

PR #421, opened 2022-08-18
Feature/refactor control

ahellander:
## Description
This is a draft PR to start a discussion on a refactor of the library structure. The goal is to move towards  a structure where the controller are more independent of the FEDn network management, and to make controllers more easily pluggable by developers and end-users. In this PR we take a first step: 

1. New module "network" which gathers reducer, combiner, controller, dashboards. 
2. Gather aggregators with combiner

Future work would be to split "dashboard" into a rest API for the "network" (add, assign, etc) and the UI (dashboard).


ahellander:
Closing since refactor-control contains this.

PR #425, opened 2022-09-21
Fix error in deployment docs

ahellander:
Fixes an error in the deployment documentation and updates the text about certificates for combiners.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -2,12 +2,10 @@ Deployment
 ======================
 
 This guide serves as reference deployment for setting up a FEDn network consisting of:
-   -  One host/VM serving the supporting services (MinIO, MongoDB)
-   -  One host/VM serving the controller / reducer 
-   -  One host/VM running a combiner 
-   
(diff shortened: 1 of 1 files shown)


docs/deployment.rst
```diff
@@ -27,9 +27,8 @@ We assume that you have root access to 3 Ubuntu 20.04 Server hosts / VMs. We rec
 Certificates (optional)
 .......................
 
-Certificates are needed for the Reducer and each of the Combiners to enable SSL for secure communication. 
-By default, FEDn will generate unsigned certificates for the reducer and each combiner using OpenSSL, but you can 
-also provide your own certificates. 
+You can provide certificates for each of the Combiners to enable secure gRPC connections. 
+FEDn can also generate unsigned certificates for using OpenSSL. The default deployment uses insecure channels, assuming that a production deployment will use a reverse proxy.
```
Wrede, line 31: Client certificates (need for self-signed user provided certs) can be specified via the env variable FEDN_GRPC_ROOT_CERT_PATH

Wrede (changes requested):

docs/deployment.rst
ahellander: I think the best thing would be if you uodated this section, since you implemented this.

ahellander:
Will improve docs further later, but for now I just want to correct the error.

Wrede (approved):

PR #443, opened 2022-11-23
Feature/436

ahellander:
## Description
Various improvements to the UI/dashboard. Moved some plots to more logical places. Hidden some others that will be moved to other places in the near future.  

1. The Model Validation box plot is now on the models view. 
2. The network view now only has the network graph and tables 
3. The dashboard view collects all the performance/tracing plots. 
4. The client histogram (statistical heterogeniety) plot has been moved to its own plot. 

Legends, axis of plots improved throughout.


Wrede (approved):
looks good

PR #465, opened 2023-05-17
Redis for backend statestore.

morganekmefjord:
## Description
Redis for backend statestore

As there are interfaces for statestore and modelstore the option to switch to redis for statestore is introduced.

### caveats/not finished
the mongotracer has been deeply integrated in several places. This should be replaced with a factory and a clean interface such that any tracer could be used for distribtuted logging. perhaps fluentd? 
The tracer has been disabled for now.

### DIFF
--- a/config/reducer-settings.override.yaml
+++ b/config/reducer-settings.override.yaml
@@ -2,8 +2,8 @@
 version: '3.3'
 
 # Overriding settings
-services:
-  reducer:
-    volumes:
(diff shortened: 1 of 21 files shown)


ahellander:
Great work! We will need to do some work to update this PR for the refactor in the upcoming 0.5 release (now in develop)

ahellander:
I will close this now to pick it up again when we work on the backend storage integrations.

PR #473, opened 2023-06-19
Combiner Report now returns the current size of the aggregator model …

ahellander:
## Description
Report the current size of the combiner model aggregation queue.


Wrede:
This is the same as #474? Think you have mixed up branches?

PR #474, opened 2023-06-19
Bug/SK-507 | Add functionality to inspect combiner queue state (debug mode)

ahellander:
## Description

Also implements feature SK-503.

### DIFF
--- a/fedn/fedn/network/combiner/round.py
+++ b/fedn/fedn/network/combiner/round.py
@@ -331,13 +331,11 @@ class RoundController:
             model_id = str(uuid.uuid4())
             self.modelservice.set_model(a, model_id)
             a.close()
+            data['model_id'] = model_id
 
-        data['model_id'] = model_id
-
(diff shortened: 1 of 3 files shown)


fedn/fedn/network/combiner/round.py
```diff
@@ -331,13 +331,11 @@ def execute_training_round(self, config):
             model_id = str(uuid.uuid4())
             self.modelservice.set_model(a, model_id)
             a.close()
+            data['model_id'] = model_id
```
Wrede, line 334: report server status if model is None? So one can debug...

Wrede (approved):
approved but with comment.

Wrede:
This is not a bug. Think you have mixed up issues IDs.

fedn/fedn/network/combiner/round.py
ahellander: This is reported on line 325.

PR #476, opened 2023-07-31
Feature/SK-505 | Flush model update queues at new session + Buffer size config

ahellander:
## Description
Update the global round logic such that the combiner aggregation queues are flushed at the beginning of every new session.  Make 'buffer size' configurable. 

Resolves 
   - SK-505
   - SK-520


fedn/fedn/network/combiner/round.py
```diff
@@ -150,16 +149,21 @@ def _training_round(self, config, clients):
...
 
+        # If buffer_size is -1 (default), the round terminates when/if all clients have completed.
+        if int(config['buffer_size']) == -1:
+            buffer_size = len(clients)
+        else:
+            buffer_size = int(config['buffer_size'])
+
         # Wait / block until the round termination policy has been met.
-        self.waitforit(config, buffer_size=len(clients))
+        self.waitforit(config, buffer_size=buffer_size)
```
Wrede, line 159: will this work with async incremental combining? the task (model update) will be task_done before new updates are added to the queue?

Wrede (changes requested):

fedn/fedn/network/combiner/round.py
ahellander: Iincremental combining is no longer the default case. We have reverted that behavior in order to accomodate more forms of aggregation. 

However I think it would still work, it will just never hit the buffer_size cap.

Wrede (approved):

PR #477, opened 2023-08-14
Feature/SK-523 | Refactor REST and add APIClient

Wrede:
## Description
This is a major refactor of the REST-API. A seperate service called api-server has been implemented. Further APIClient has been added which can be imported by:
`fedn import APIClient`
`client = APIClient(host="localhost", port=8092)`

With the client you can query the statestore and send requests to the controller. e.g upload package, initial models, download package, list clients, list validations, list events etc. 

To consider before merge:
-  should the dashboard be removed/relocated?
- connect.py and package.py has been moved to its corresponding service where they are used from the "common" folder. Should old one be removed?
- POST endpoints include json where key, value will be sent at kwargs to logic functions, i.e add_combiner, add_client and start_session. This means that any client sending a request need to have json keys corresponding to the arguments in the API calss functions. 
- get_reducer and set_reducer is no longer used except for in the old dashboard...remove?

## Extras
Add docsstring is almost complete for the whole network package. SDK should be added to sphinx docs. The Client class have many private functions (_) which makes the SDK doc very sparse for this class! Loadbalancer have two files with the same class names!


fedn/fedn/network/api/client.py
```diff
@@ -0,0 +1,263 @@
...
+
+        :param round_id: The round id to get.
+        :type round_id: str
+        :return: The round config and metrics.
+        :rtype: dict
+        """
+        response = requests.get(self._get_url(f'get_round?round_id={round_id}'), verify=self.verify)
+        return response.json()
+
+    def start_session(self, session_id=None, round_timeout=180, rounds=5, round_buffer_size=-1, delete_models=True,
```
ahellander, line 101: It seems to not take in the parameters correctly, not that I here ser rounds and helper, but they were ignored: 

>>> client.start_session(round_timeout=30,rounds=1,helper='pytorchhelper')
{'config': {'clients_requested': 1, 'clients_required': 1, 'committed_at': '2023-08-16 05:58:29', 'delete_models_storage': True, 'helper_type': 'kerashelper', 'model_id': 'a1043100-e918-4855-a240-09a360ea76ca', 'round_timeout': 180.0, 'rounds': 5, 'session_id': '1fa4cd6d-53cd-41a6-b66f-ea6e609bd755', 'task': '', 'validate': True}, 'message': 'Session started successfully.', 'success': True}

ahellander (changes requested):
Great progress!

ahellander (commented):
Looks great! We can merge this and iterate from there.

ahellander (approved):

PR #479, opened 2023-10-03
Feature/SK-545 | Statestore init from api-server lack proper setup

Wrede:
When running api-server alone without dahsboard the statestore was not properly configured due to the "default" arg for the mongostatestore constructor. The "default" arg has been removed since we rarely use control settings from file (instead send JSON to start session endpoint).


ahellander (approved):
Looks great!

PR #484, opened 2023-11-13
Feature/SK-558 | Adds Python's logging framework to the client

stefanhellander:
Adds logging using Python's logging framework to the client, instead of relying on print statements. Also fixes the option to log to file. You can set the verbosity like this:

fedn run client --init client.yaml --verbosity DEBUG

Default verbosity is INFO.


Wrede (changes requested):
Nice! If you merge in develop into this feature branch I think you will have some regular prints that you can replace with logging as well.

ahellander (commented):
Great! Would love the same in the rest of the code base as well :-)

Wrede (approved):

PR #487, opened 2023-11-15
Feature/SK-583 | Bump grpcio and grpcio-tools

stefanhellander:
## Description
Bump grpcio and grpcio-tools to latest versions. Update setup.py to support 3.11, 3.12, removed 3.7.


Wrede:
I know grpcio 1.57 tensorflow-cpu==2.13.1 and pytorch 1.13. But I think I had problems with 1.58, might be newer versions of tf and torch though that works.

Wrede (changes requested):
Fedn installs protobuf 4.25.0 as a dependency of grpcio 1.59. However, when you install tensorflow 2.9 it downgrades protobuf to 3.19.6:

#13 36.36   Attempting uninstall: protobuf
#13 36.36     Found existing installation: protobuf 4.25.0
#13 36.36     Uninstalling protobuf-4.25.0:
#13 36.37       Successfully uninstalled protobuf-4.25.0
#13 47.92 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
#13 47.92 grpcio-tools 1.59.2 requires protobuf<5.0dev,>=4.21.6, but you have protobuf 3.19.6 which is incompatible.

Tensorflow need to be updated (see my previous comment). Not sure TF supports 1.59.

Wrede:
Closes fix en #490

PR #494, opened 2023-12-12
Feature/Android

niklastheman:
Android helper added


fedn/fedn/utils/plugins/androidhelper.py
```diff
@@ -0,0 +1,108 @@
...
+class Helper(HelperBase):
+    """FEDn helper class for android json model weights."""
+
+    def __init__(self):
+        """Initialize helper."""
+        self.name = "androidhelper"
+        super().__init__()
+
+    # function to calculate an incremental weighted average of the weights
+    def increment_average(
```
ahellander, line 19: Why do we need two version of this function? The implementation in increment_average_add looks more efficient (vectorized). Why not have only that one (as 'increment_average')?

fedn/fedn/utils/plugins/androidhelper.py
```diff
@@ -0,0 +1,108 @@
+import json
+import os
+import tempfile
+
+import numpy as np
+
+from .helperbase import HelperBase
+
+
+class Helper(HelperBase):
```
ahellander, line 10: Maybe more a philosophical question, but there seems to be nothing here specific to Android. It is more that we are here assuming that the model parameters have been serialized to a json document, and that that document can be cast to a numpy.array? Maybe add those assumptions to the readme?  I guess what I am after is if it should be named "androidhelper" or something else (jsonhelper?)

ahellander (changes requested):

fedn/fedn/utils/plugins/androidhelper.py
niklastheman: Think this is very much still in dev state. Will convert to draft.

ahellander:
@niklastheman This one should be closed now, right? You merged the android helper in a separate PR, right?

niklastheman:
> @niklastheman This one should be closed now, right? You merged the android helper in a separate PR, right?

Yes, this is obsolete.

niklastheman:
Has been solved in other PR.

PR #497, opened 2023-12-28
Make sure combiner starts before clients in the pseudo-distributed sandbox 

### LINKED ISSUE #496: When restarting the pseudo-distributed sandbox the clients do not find the combiner
If you start the example sandbox as such: 

docker-compose -f ../../docker-compose.yaml -f docker-compose.override.yaml up 

Then stop, and restart, the clients do not manage to attach to the combiner (now set in the db). 

Workaround: Flushing the db before restart. 

How if should work: The default combiner name is "combiner" and the clients should manage to get assigned to this upon restart. 

**Severity**
- [ ] Critical/Blocker (select if the issue makes the application unusable or causes serious data loss)
- [ ] High (select if the issue affects a major feature and there is no workaround or the available workaround is very complex)
- [x] Medium (select if the issue affects a minor feature or affects a major feature but has an easy enough workaround to not cause any major inconvenience)
- [ ] Low (select if the issue doesn't significantly affect the user experience, like minor visual bugs)

**Describe the bug**

**Environment:**
 - OS: [e.g. mac OS]
 - Version: [e.g. macOS Catalina]

**Reproduction Steps**

**Expected behavior**

**Screenshots**

**Contact Details**
An e-mail address in case we need to contact you for further details regarding this issue.

ahellander:
## Description

Closes #496

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -136,3 +136,5 @@ services:
       - "/venv/bin/pip install --no-cache-dir -e /app/fedn && /venv/bin/fedn run client --init config/settings-client.yaml"
     deploy:
       replicas: 0
+    depends_on:
+      - combiner


ahellander:
This actually does not fix the problem. I think we need/should implement healthchecking for the combiner server.

PR #515, opened 2024-02-03
Feature/SK-658 | Add gRPC healthchecking

ahellander:
## Description

This PR introduces the standardized gRPC HealthChecking protocol for the combiner server. 

- Add HealthCheck servicer
- Add grpc_health_probe to Docker container image
- Healthcheck on combiner service in docker-compose
- Ensures that combiner is ready before clients start  

This fixes the issue that intermittently when spinning up the pseudo-distributed sandbox with docker-compose ,the clients fail with UnMatchedConfig error due to the combiner not being ready yet. 

We can also now use this for Kubernetes: https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/


Dockerfile
```diff
@@ -13,7 +13,13 @@ COPY config/settings-reducer.yaml.template /app/config/settings-reducer.yaml
 COPY $REQUIREMENTS /app/config/requirements.txt
 
 # Install developer tools (needed for psutil)
-RUN apt-get update && apt-get install -y python3-dev gcc
+RUN apt-get update && apt-get install -y python3-dev gcc wget
```
Wrede, line 16: installing binary wget not good, remove executable after it has been used. (we should have a build layer in the image really)

Dockerfile
```diff
@@ -13,7 +13,13 @@ COPY config/settings-reducer.yaml.template /app/config/settings-reducer.yaml
...
+
+# Install grpc health probe checker
+RUN GRPC_HEALTH_PROBE_VERSION=v0.4.24 && \
```
Wrede, line 19: Ok, but we won't need this binary for k8s. It's good for compose but other than that it just add more maintenance. For example, this version maybe need to be in sync with grpcio version.

Wrede (changes requested):
Nice! but there are something that need to be considered...

Wrede:
one possible solution is to introduce an ARG health_checking=0 in the dockerfile. Then download probe only if health_checking=1. Then we set this arg in compose file to be 1. I.e by default it will not be in the image, but the healthcheck endpoint will still be there for clients to use in what ever way they desire.

Wrede (approved):

PR #516, opened 2024-02-03
Feature/refactor proto

ahellander:
## Description

We hade multiple servicers defined in the proto (connector, control, reducer),  that is in the end all implemented in Combiner.  I think it makes the code more understandable if we merge them in combiner, and then if we want we can refactor the server in the future.


ahellander:
Contains SK-658 so that one should be merged first.

PR #520, opened 2024-02-05
Feature/SK-671 | Merge services in combiner, proto refactor

Wrede:
closes PR #516 (remake PR)

We hade multiple servicers defined in the proto (connector, control, reducer), that is in the end all implemented in Combiner. I think it makes the code more understandable if we merge them in combiner, and then if we want we can refactor the server in the future.


ahellander (approved):
Looks good, just fix the conficting grpc files.

Wrede:
Close due to inactivity, might come back to this at a certain point

PR #523, opened 2024-02-06
Bugfix/SK-675 | Model staging not in sync

Wrede:
Issue with model artifacts not being present at combiner disk during validation (clients request download validation model). We think the issue is related to latency between training and validations requests. That training is for some reason faster than validation and deletes the staged model which is waiting to be validated. 

Our solution to this is to merge validation stream and training stream into a single TaskStream which are then differentiated by a message field "type" of enum StatusType in the grpc TaskRequest.


ahellander (approved):
Looks good, improves robustness.

PR #526, opened 2024-02-08
Feature/SK-623 | Validations objects should store proper timestamp

niklastheman:
validation timestamp type change: string => timestamp


ahellander (approved):
Good stuff.

PR #527, opened 2024-02-08
Feature/SK-684 | Events should store proper timestamps

niklastheman:
status timestamp type change: string => timestamp


ahellander (approved):
Nice.

PR #530, opened 2024-02-16
Docs/SK-691 | Add tutorial for distributed deployment

viktorvaladi:
Adds a tutorial to the docs for distributed deployment.

### DIFF
--- a/docs/distributed.rst
+++ b/docs/distributed.rst
@@ -6,7 +6,7 @@ the host and different devices as clients. For general steps on how to run FEDn,
 
 
 .. note::
-   For a secure and production-grade deployment solution over **public networks**, explore the FEDn studio service at 
+   For a secure and production-grade deployment solution over **public networks**, explore the FEDn Studio service at 
    **studio.scaleoutsystems.com**. 
    
    Alternatively follow this tutorial substituting the hosts local IP with your public IP, open the neccesary 
--- a/docs/distributed.rst
+++ b/docs/distributed.rst
@@ -92,7 +92,12 @@ build a custom Docker image:
 Configuring and Attaching Clients
 -------------
 
-The next step is to configure and attach clients. For this we need to download data and make data partitions: 
+The next step is to configure and attach clients. Begin with navigating to ``examples/mnist-pytorch`` and 
+execute the initializion script:
+
+.. code-block:: python
+
+   bin/init_venv.sh
 
 Download the data:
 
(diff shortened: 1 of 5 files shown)


ahellander (changes requested):
Nice work. However, I was more envisaging a more detailed step-by-step tutorial showing how to configure a RPI4 with FEDn and starting it as a client, probably connecting it to an existing FEDn deploy and FEDn Studio.

ahellander:
So I think it will be more pedagogical to split in two: 

1. Distributed deployment (docs) 
2. Example / tutorial with Rasberry PI5.

viktorvaladi:
> So I think it will be more pedagogical to split in two:
> 
> 1. Distributed deployment (docs)
> 2. Example / tutorial with Rasberry PI5.

Sounds good! I will fix that.

docs/distributed.rst
```diff
@@ -0,0 +1,67 @@
...
+-  `One host workstation and atleast one client device`
+-  `Python 3.8, 3.9 or 3.10 <https://www.python.org/downloads>`__
+-  `Docker <https://docs.docker.com/get-docker>`__
+-  `Docker Compose <https://docs.docker.com/compose/install>`__
+
+Launch a distributed FEDn Network 
+-------------
+
+
+Note the local IP address of your host device; it will be needed later. Follow the standard procedure
```
ahellander, line 28: Better here to replicate how you start up the server-side? Suggest the user to use docker-compose? Also mention where to find the ports they need to open?

docs/distributed.rst
```diff
@@ -0,0 +1,67 @@
...
+   —add-host=api-server:<host local ip> \
+   —add-host=combiner:<host local ip> \
+   <image name> run client -in client.yaml --name client1
+
+
+Alternatively updating the `/etc/hosts` file, appending the following lines for running naitively:
+
+.. code-block::
+
+   <host local ip>      api-server
```
ahellander, line 60: Should be public ip, right?

ahellander (commented):
Also, I suggest that you highlight that this is not a secure deployment (unless the users takes separate measures) and that you refer to FEDn Studio for a production-grade deployment.

docs/distributed.rst
viktorvaladi: This is mainly for deployment over local networks, so the host's local network IP. I've clarified that and referred to Studio for secure deployment.

docs/distributed.rst
viktorvaladi: I suggested docker-compose and added where to find the ports in the note.

docs/distributed.rst
ahellander: If it is a distributed setup you cannot communicate on the hosts local IP from a client. I think the setting that all machines are on the same local network is fairly uncommon.

ahellander: Ok, I see, you made that clear in the beginning of the tutorial.

ahellander (approved):

PR #533, opened 2024-02-23
Feature/SK-697 | Adds token auth to the APIClient

stefanhellander:
## Description
Adds the option to pass an auth token when initializing the APIClient (either as input to the constructor or by setting the environment variable FEDN_AUTH_TOKEN).

### DIFF
--- a/fedn/fedn/network/api/client.py
+++ b/fedn/fedn/network/api/client.py
@@ -1,6 +1,10 @@
 import json
+import os
 
 import requests
+
+from fedn.common.log_config import (logger, set_log_level_from_string,
+                                    set_log_stream)
 
 __all__ = ['APIClient']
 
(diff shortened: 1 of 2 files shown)


fedn/fedn/network/api/client.py
```diff
@@ -18,26 +19,36 @@ class APIClient:
...
+    def __init__(self, host, port=None, secure=False, verify=False, token=None, verbosity="INFO"):
         self.host = host
         self.port = port
         self.secure = secure
         self.verify = verify
+        self.header = {}
+        # Check if auth token is set by environment variable
+        env_token = os.environ.get("FEDN_AUTH_TOKEN", False)
+        if env_token:
+            self.header = {"Authorization": "Token {}".format(token)}
```
Wrede, line 31: The auth scheme should also be configurable. e.g via env var FEDN_AUTH_SCHEME, where default can be "Token" but for JWTs its "Bearer".

Wrede (changes requested):

Wrede (approved):

PR #534, opened 2024-02-26
Feature/SK-544 | Server side changes for c++ client

Wrede:
Introduces protobuf changes. pakage is now called "fedn" instead of "grpc" (caused conflicts in protoc c++ lib). Added field "config" to ModelUpdate message instead of processing nested json in "meta" field. Added changes to AggregatorBase to still allow the nested json and the new field.


fedn/fedn/network/combiner/aggregators/aggregatorbase.py
```diff
@@ -116,11 +116,17 @@ def load_model_update(self, model_update, helper):
...
         # Get relevant metadata
-        data = json.loads(model_update.meta)['training_metadata']
-        config = json.loads(json.loads(model_update.meta)['config'])
-        data['round_id'] = config['round_id']
-
-        return model, data
+        metadata = json.loads(model_update.meta)
+        if 'config' in metadata.keys():
+            # Used in Python client
+            config = json.loads(metadata['config'])
```
ahellander, line 122: Why do we need to support both versions of the message? Why not just introduce the new config?

Wrede: because of backward compability

ahellander (approved):

PR #535, opened 2024-02-26
Feature/SK-702 | Base-services config by Envs

Wrede:
It should be possible to setup connection configurations to mongo and minio via ENVs.


ahellander (changes requested):

Wrede:
Close due to inactivity

PR #537, opened 2024-02-28
Feature/SK-707 | Flower client example

viktorvaladi:
Adds an example and readme for how you can implement a flower client in fedn.


examples/flower-client/client/entrypoint
```diff
@@ -0,0 +1,121 @@
...
+    # Figure out FEDn client number from container name
+    number = os.environ.get("CLIENT_NUMBER", "0")
+    return number
+
+
+def _get_flower_client():
+    """Instanziates flower client."""
+    original_argv = sys.argv
+    # Set sys.argv to mimic command-line input required by flower_client.py
+    sys.argv = ["flower_client.py", "--node-id", _get_node_id()]
```
ahellander, line 25: I don't get why this is needed? Can you explain a bit?

examples/flower-client/client/flwr_client.py
```diff
@@ -0,0 +1,138 @@
+import argparse
```
ahellander, line 1: Add attribution that this comes from the Flower repo, and provide a link back there.

examples/flower-client/client/flwr_client.py
```diff
@@ -0,0 +1,138 @@
...
+        train(net, trainloader, epochs=1)
+        return self.get_parameters(config={}), len(trainloader.dataset), {}
+
+    def evaluate(self, parameters, config):
+        self.set_parameters(parameters)
+        loss, accuracy = test(net, testloader)
+        return loss, len(testloader.dataset), {"accuracy": accuracy}
+
+
+if __name__ == "__main__":
```
ahellander, line 133: This is never used when the example is wrapped by FEDn, right?

examples/flower-client/requirements.txt
```diff
@@ -0,0 +1,6 @@
+torch==1.13.1
+torchvision==0.14.1
+fire==0.3.1
+docker==6.1.1
```
ahellander, line 4: docker not needed ?

examples/flower-client/client/entrypoint
ahellander: We don't need the actual Flower-client running, right?

ahellander (changes requested):

examples/flower-client/client/entrypoint
viktorvaladi: Thats true, dont need to run the Flower-client, running it complicates the process since It's gets wrapped up in custom flower types that would need to be handled. 

The reason for this code is that I did not want to change anything in the flower file, and it depends on system args. With our code using Fire it's a clash with sys args. So my approach was to programmatically add system args before importing flower, and then resetting back to our Fire args afterwards.

examples/flower-client/client/flwr_client.py
viktorvaladi: True. I added this part to the flower example so that the client would not run..

examples/flower-client/client/entrypoint
ahellander: I see, ok. Still don't fully get why it would depend on some systems args if we do not make any systems call?

viktorvaladi: The flower code example expects the code the be run with "--node-id" added as an argument. I can work around it as well but then I would need to alter the flower client file a little

ahellander: So the thing to do seems to be to define the input in a config dict instead.

ahellander (approved):
Looks good!

examples/flower-client/client/flwr_task.py
```diff
@@ -92,4 +92,4 @@ def get_weights(net):
 def set_weights(net, parameters):
     params_dict = zip(net.state_dict().keys(), parameters)
     state_dict = OrderedDict({k: torch.tensor(v) for k, v in params_dict})
-    net.load_state_dict(state_dict, strict=True)
\ No newline at end of file
+    net.load_state_dict(state_dict, strict=True)
```
ahellander, line 95: Is this from the Flower example or did we add it? We should have unmodified flwr_task.py

ahellander: good

PR #541, opened 2024-03-11
Feature/async simulation

ahellander:
## Description

Example of working with asynchronous clients in FEDn


sztoor (approved):
Seem correct.

PR #543, opened 2024-03-11
Feature/SK-721 | Add JWT verification (decoding) 

Wrede:
Adds authentication decorators using JWT. Introducing RBAC with custom claims "role" in issued tokens. Current roles are:
- "client"
- "combiner"
- "admin"

If env FEDN_JWT_SECRET_KEY is set, JWT verification are enabled. 
Custom claims can be added by setting FEDN_JWT_CUSTOM_CLAIM_KEY and verified against FEDN_JWT_CUSTOM_CLAIM_VALUE. 

FEDN_AUTH_SCHEME, default to "Token".

OBS that FEDn do not issue the tokens.


fedn/fedn/network/api/client.py
niklastheman, line 1: auth_scheme = os.environ.get("FEDN_AUTH_SCHEME", "Bearer"), maybe set this to Token?

niklastheman (approved):

niklastheman (approved):

PR #544, opened 2024-03-13
Feature/SK-710 | FEDn cli - init server

niklastheman:
# FEDn cli

_Run FEDn, list entities, start sessions etc. Default to running commands for localhost but can be configured, either by params or ENV:s._

**Config includes:**
- FEDN_PROTOCOL - communication protocol (default http)
- FEDN_HOST - controller host (default localhost)
- FEDN_PORT - controller port (default 8092)
- FEDN_TOKEN - token (used for authentication in FEDn Studio)
- FEDN_AUTH_SCHEME - token auth scheme (Bearer/Token) 

## Added commands:

**fedn run**
- start controller and combiner

**fedn list <entity>**
- list entities (models, packages, clients etc)

**fedn start session**
- start a new session

**fedn upload <type>**
- upload compute package or seed model

**fedn config**
- see currently set ENV:s used as config
- _fedn config generate_ can be used to generate script to set the ENV:s
- _fedn config clear_ can be used to generate script to clear the ENV:s

### DIFF
--- a/examples/mnist-pytorch/client/entrypoint
+++ b/examples/mnist-pytorch/client/entrypoint
@@ -3,7 +3,6 @@ import collections
 import math
 import os
 
-import docker
 import fire
 import torch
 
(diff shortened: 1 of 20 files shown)


ahellander (commented):
Looks good overall imo. 

However, I wonder if we should consider refactoring the command naming scheme in relation to SK-746 (aka "python_env"). 

One model would be to follow the MLFlow Project convention to use "run" to exectute entry points as specified in fedn.yaml. 

In that case we could perhaps do

fedn client start, stop 
fedn combiner start, stop
fedn list 

?

niklastheman:
> Looks good overall imo.
> 
> However, I wonder if we should consider refactoring the command naming scheme in relation to SK-746 (aka "python_env").
> 
> One model would be to follow the MLFlow Project convention to use "run" to exectute entry points as specified in fedn.yaml.
> 
> In that case we could perhaps do
> 
> fedn client start, stop fedn combiner start, stop fedn list
> 
> ?

I think if it harmonises better with MLFlow it's a good idea. We should bring @Wrede into the discussion.

PR #545, opened 2024-03-14
Feature/SK-724 | API method to load a FEDn model / parameters from S3.

niklastheman:
# Download model file and its parameters

## Two new endpoints added

**GET /models/_model-id_/download**
- downloads the model file
- id can be either the model property of the model object or the document id

**GET /models/_model-id_/parameters**
-  gets the parameters of the model
- id can be either the model property of the model object or the document id
- please note this only works for models that are stored as numpy arrays

### DIFF
--- a/fedn/fedn/network/api/v1/model_routes.py
+++ b/fedn/fedn/network/api/v1/model_routes.py
@@ -1,3 +1,6 @@
+import io
+
+import numpy as np
 from flask import Blueprint, jsonify, request, send_file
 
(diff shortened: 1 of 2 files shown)


ahellander (commented):
Looks great and what I need, but I also need wrappers in the APIClient :-)

niklastheman:
> Looks great and what I need, but I also need wrappers in the APIClient :-)

Yes, I will fix this but would like to do it in a separate PR if okey with you? The APIClient needs to be refactored in general.

ahellander (approved):

PR #549, opened 2024-03-18
Docs/add studio to readme

ahellander:
## Description
Add more information about FEDn Studio to main readme.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,8 +26,6 @@ Core Features
 -  **Track events and training progress in real-time**. Extensive event logging and distributed tracing helps developers monitor experiments in real-time, facilitating troubleshooting and auditing.  
    Tracking and model validation data can easily be retrieved using the API enabling development of custom dashboards and visualizations. 
 
--  **Robust in asynchronous federated learning scenarios**. FEDn handles clients that connects and disconnects during training. 
-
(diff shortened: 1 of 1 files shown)


sztoor (commented):
Overall, good and clear description. 

For the part "Develop a FEDn project in a local development environment, and then deploy it to FEDn Studio" 

We can make two branches: 

1 - FL Data Scientist path -> All-in-one to more FEDn SDK
2 - FL Production FL -> All-in-one to FEDn Studio

Strong points of this approach:
+  Will show the depth
+  Make it more clear based on the different roles 

Weak points of this approach:
- May be confusing and lengthy

niklastheman (approved):
I think it's a good improvement!
A couple of thoughts though:

- Maybe we should get rid of the top logo in favor of the new one (we could perhaps "draw" it our selfs in the README to make it look "tech")? Also the logos (contributions) in the bottom could be better arranged and the Scaleout one is old...
-  Do we need to mention Pytorch in the Quickstart section or could it simply say Getting started or Quickstart? Also I am still not thrilled about the term SDK. For med FEDn is first and foremost a framework.
- The link to the FEDn Studio docs will change soon. Don't have to wait for it though...

PR #556, opened 2024-03-26
Feature/SK-746 | Add python_env support in fedn.yaml

Wrede:
Introduces python_env in fedn.yaml to control python environments. 

Similar to MLFlow projects one can define a python_env: py_env.yaml which will be processed by fedn client and creates a virtual environment via virtualenv package. Currently there is no support to handle different python versions, the the aim is to implement this in a future release using pyenv package. 

There's also a new _exec_cmd function to handle subprocesses, which support both unix and non-unix OS.

You can now set package excract directy with env FEDN_PACKAGE_EXTRACT_DIR

Only mnist-pytorch example has been updated with a python environment using the new feature. You will now be able to start the client using the mnist-pytorch example in the follwoing way:

pip install fedn
fedn run client -in client.yaml

This PR also introduces "fedn package create" and "fedn run build" in the CLI. "package" will make a tarball of a fedn project directory and "run build" will execute an entry_point in fedn.yaml defined under "build" key. This can for example be used to build initial model artifacts.

The package is built with examples/mnist-pytorch/bin/build.sh
Training data and seed is genereated using the "startup" entry_point, instead of doing this manually. 

Docs has not been updated yet!!


niklastheman (approved):

examples/mnist-pytorch/README.rst
```diff
@@ -23,108 +28,96 @@ Clone this repository, locate into this directory:
...
 
 .. code-block::
 
-   bin/init_venv.sh
-
-Next create the compute package and a seed model:
+   pip install fedn 
+   export FEDN_AUTH_SCHEME=Bearer
+   export FEDN_PACKAGE_EXTRACT_DIR=package
+   fedn run client -in client.yaml --name client1 --secure=True --force-ssl
```
ahellander, line 48: I think if there is a name set in the client.yaml, the cli flag does not take effect?

examples/mnist-pytorch/README.rst
```diff
@@ -23,108 +28,96 @@ Clone this repository, locate into this directory:
...
 
-.. code-block::
 
-   bin/build.sh
+Upload the initial model to FEDn Studio project. The seed.npz file is created when you run start the client and is found in ./data/models/seed.npz
```
ahellander, line 51: I understand that the intention is to make it "one-shot" to start the example, but I worry that it would feel a bit illogical that the seed is created by the client upon startup? It feels that that step belongs to. a "model initiator" part of the flow?

examples/mnist-pytorch/README.rst
```diff
@@ -23,108 +28,96 @@ Clone this repository, locate into this directory:
...
+   git clone https://github.com/scaleoutsystems/fedn.git
+   cd fedn/examples/mnist-pytorch
 
-Split the data in 10 partitions:
+Start a pseudo-distributed FEDn network using docker-compose:
 
 .. code-block::
 
-   bin/split_data --n_splits=10
+   docker compose \
```
ahellander, line 70: For a user not super-familiar with docker, I think it helps to see the terminal output when starting up the network.

ahellander: Is it because you want them to stay in the same terminal?

examples/mnist-pytorch/README.rst
```diff
@@ -23,108 +28,96 @@ Clone this repository, locate into this directory:
...
 
-   docker run \
-  -v $PWD/client.yaml:/app/client.yaml \
-  -v $PWD/data/clients/1:/var/data \
-  -e ENTRYPOINT_OPTS=--data_path=/var/data/mnist.pt \
-  --network=fedn_default \
-  ghcr.io/scaleoutsystems/fedn/fedn:master-mnist-pytorch run client -in client.yaml --name client1
+   bin/build.sh
+
+You should now have a file 'package.tgz'.
```
ahellander, line 96: For me it has always felt "wrong" to attach clients before the network/project has been initialized.

examples/mnist-pytorch/README.rst
```diff
@@ -23,108 +28,96 @@ Clone this repository, locate into this directory:
...
-  -v $PWD/data/clients/1:/var/data \
-  -e ENTRYPOINT_OPTS=--data_path=/var/data/mnist.pt \
-  --network=fedn_default \
-  ghcr.io/scaleoutsystems/fedn/fedn:master-mnist-pytorch run client -in client.yaml --name client1
+   bin/build.sh
+
+You should now have a file 'package.tgz'. 
 
-Observe the API Server logs and combiner logs, you should see the client connecting and entering into a state asking for a compute package. 
+You are now ready to use the API to initialize the system with the package.
```
ahellander, line 98: Add instructions for how to do that inline here. Or add link to the right place in docs.

ahellander (changes requested):

ahellander (approved):

PR #558, opened 2024-03-31
Update README.rst

ahellander:
Improve main readme, highlight option to use SaaS for dev/test.


sztoor (commented):
Overall looks good. How about if we also add one more like "With our SaaS platform, users don't have to worry about the server-side components. The local computational resources will be used for client computations..... " Something more to highlight user benefits.

PR #565, opened 2024-04-08
Bugfix/sk 752

ahellander:
## Description

Bugfix - missing self.attach() upon client reconnection attempt. 

Also introduces the logic change that if the reason for a client being disconnected is Token expirations, no reconnection attempts are made and the client exits. 

Closes SK-752.

### DIFF
--- a/fedn/fedn/network/clients/client.py
+++ b/fedn/fedn/network/clients/client.py
@@ -773,7 +773,7 @@ class Client:
             cnt = 0
             old_state = self.state
             while True:
-                time.sleep(5)
+                time.sleep(1)
                 if cnt == 0:
                     logger.info("Client is active, waiting for model update requests.")
                     cnt = 1
--- a/fedn/fedn/network/clients/client.py
+++ b/fedn/fedn/network/clients/client.py
@@ -773,16 +773,18 @@ class Client:
             cnt = 0
             old_state = self.state
             while True:
-                time.sleep(1)
+                time.sleep(5)
                 if cnt == 0:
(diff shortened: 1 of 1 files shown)


fedn/fedn/network/clients/client.py
```diff
@@ -709,14 +703,18 @@ def _send_heartbeat(self, update_frequency=2.0):
...
+                    if self._missed_heartbeat > self.config['reconnect_after_missed_heartbeat']:
+                        self.disconnect()
                 if status_code == grpc.StatusCode.UNAUTHENTICATED:
                     details = e.details()
                     if details == 'Token expired':
-                        logger.warning("GRPC hearbeat: Token expired. Reconnecting.")
-                        self.detach()
+                        logger.error("GRPC hearbeat: Token expired. Disconnecting.")
+                        self.disconnect()
+                        sys.exit("Unauthorized. Token expired. Please obtain a new token.")
```
Wrede, line 716: Why? this will break the client trying to refresh token?

Wrede (changes requested):
See comment.

fedn/fedn/network/clients/client.py
ahellander: Yes, was thinking about that, but currently the client is not trying to refresh the token, right? So for now the right things seems to be to disconnect?

Wrede: no that's not right they will try to refresh tokens, if you ask for new assignment and the token has expired it will try to get a new one. But only when calling REST-API

Wrede (approved):

PR #584, opened 2024-04-18
using dataloader in mnist-pytorch example

mattiasakesson:


ahellander (changes requested):
Getting this error when the client starts to train: 

Screenshot 2024-04-18 at 23 31 00

PR #587, opened 2024-04-22
Bugfix/SK-820 | Add missing / to some api paths

Wrede:
Fixes some endpoints not working with Studio due to incorrect redirects.


ahellander (approved):
Tested locally, it works.

PR #591, opened 2024-04-24
Bug/SK-821 | Client status sync issue

niklastheman:
test to fix client sync issues


ahellander (approved):
Like we discussed, we should take a holistic look at this overall logic, but I think this is a good test to see if our hypothesis holds.

PR #593, opened 2024-04-26
Feature/SK-805 | FEDn cli - new order

niklastheman:
# Some changes to the CLI

- **fedn run combiner** -> **fedn combiner start**
- **fedn run client** -> **fedn client start**
- **fedn client/combiner/model/package/round/session/status/validation list** - list objects from statestore
- **refactoring** - refactored code into separate files
- **fedn config** - lists env:s (and values) relevant for the cli


docs/faq.rst
```diff
@@ -19,6 +19,17 @@ However, during development of a new model it will be necessary to reinitialize.
 
    2. Restart the clients. 
 
+Q: Can I skip fetching the remote package and instead use a local folder when developing the compute package
```
ahellander, line 22: I think I removed this Q in current  master (since --remote False flag currently not working)

ahellander (approved):
It looks right to me. I think you will have some conflicts when you merge in master though, so need to check that. 

I think it would be good if the "config" commmand also has one for actually configuring. I.e. prompt user for key configurations. But that could be a separate PR perhaps.

niklastheman:
> It looks right to me. I think you will have some conflicts when you merge in master though, so need to check that.
> 
> I think it would be good if the "config" commmand also has one for actually configuring. I.e. prompt user for key configurations. But that could be a separate PR perhaps.

- Merged
- I actually tried creating a command to set the env:s interactivity but found no way to make it work. The best solution I found was to prompt the user and the return the strings to paste in to a terminal. This felt sub optimal though so left it out. If you want it I can easily re-implement it.

Wrede (changes requested):
Looks nice! Just think we should add a "obsolete" warning or error to the old "fedn run client/combiner", since this is a major change and will break many container images. Best would be to actually allow  "run client/combiner" for a while and instead pass all arguments to the new click command, that way we can have a warning between two major releases, and then in the future update to am error.

niklastheman:
> Looks nice! Just think we should add a "obsolete" warning or error to the old "fedn run client/combiner", since this is a major change and will break many container images. Best would be to actually allow "run client/combiner" for a while and instead pass all arguments to the new click command, that way we can have a warning between two major releases, and then in the future update to am error.

deprecation warning added

Wrede (approved):
Hold?

niklastheman:
> Hold?

Sure! 👌 Can be merged prior to next release.

PR #597, opened 2024-05-06
Bug/SK-793 | Fix package extract and dispatch folder

Wrede:
on package extract (untar) tarball will now extract to default folder "package" from client CWD. The directory can be chnaged using FEDN_PACKAGE_EXTRACT_DIR environment var. 

When a package is extracted FEDn searches for "fedn.yaml" in the root and child folders in the extracted folder. The path of fedn.yaml is saved and will be the "run path" for the dispatcher. In other words relative path in entry_points will be realative "run path"


ahellander (approved):
I think it looks good. Only question is what happens if a user already has a folder called "package" in their project structure (unlikely, but could happen).

PR #606, opened 2024-05-15
bugfix/SK-850 | Sets client status to online in database when client connects

stefanhellander:
## Description
If a client disconnects and connects again before combiner registers it as offline, then controller updates status is db to "available", but combiner sees no change and client status is then stuck as "available" in database.

If a client and combiner goes down at a similar time, then the client is always listed as "online" if the combiner didn't have time to register the client disconnecting.

This PR fixes the first issue by setting the db status to online in TaskStream when the client reconnects. It partly addresses the second issue by cleaning up the status of previously connected clients when the combiner comes online again. Of course, if the combiner remains offline, then this issue is still there (could maybe be solved in the controller in that case)?


Wrede:
working on review...don't merge

Wrede (approved):

PR #608, opened 2024-05-17
Feature/SK-852 | Add numpy raw binary helper

Wrede:
The default numpyhelper is based on numpy compressed file format (npz). This PR introduces a new helper that enables raw binary file format (.bin) which will be read using either numpy.frombuffer or numpy.fromfile instead of numpy.load. 

The helper is called "binaryhelper".

### DIFF
--- a/fedn/utils/helpers/plugins/binaryhelper.py
+++ b/fedn/utils/helpers/plugins/binaryhelper.py
@@ -1,0 +1,16 @@
+from fedn.utils.helpers.plugins.numpyhelper import Helper
+
+
+class Helper(Helper):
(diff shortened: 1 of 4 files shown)


fedn/utils/helpers/plugins/binaryhelper.py
```diff
@@ -0,0 +1,16 @@
+from fedn.utils.helpers.plugins.numpyhelper import Helper
```
ahellander, line 1: Would it improve readability to do: 

from fedn.utils.helpers.plugins.numpyhelper import Helper as NumpyHelper 
..
class Helper(NumpyHelper) 

?

ahellander (approved):
Looks good! See  comment / question.

PR #614, opened 2024-05-24
Feature/SK-866 | Inference TaskType workflow

Wrede:
introduces a new TaskType and workflow for inference on the server side. 
Currently infernce will work very similar to validation except the number of rounds will only be 1. 

It's up to the client side to respond to the TaskStream when statusType  is of type "INFERENCE".

This PR does not introduce any client side inference actions.

### DIFF
--- a/fedn/network/combiner/aggregators/aggregatorbase.py
+++ b/fedn/network/combiner/aggregators/aggregatorbase.py
@@ -86,9 +86,11 @@ class AggregatorBase(ABC):
         :return: True if the model update is valid, False otherwise.
         :rtype: bool
         """
-        data = json.loads(model_update.meta)["training_metadata"]
-        if "num_examples" not in data.keys():
-            logger.error("AGGREGATOR({}): Model validation failed, num_examples missing in metadata.".format(self.name))
+        try:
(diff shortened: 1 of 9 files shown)


fedn/network/combiner/aggregators/aggregatorbase.py
```diff
@@ -86,9 +86,11 @@ def _validate_model_update(self, model_update):
         :return: True if the model update is valid, False otherwise.
         :rtype: bool
         """
-        data = json.loads(model_update.meta)["training_metadata"]
-        if "num_examples" not in data.keys():
-            logger.error("AGGREGATOR({}): Model validation failed, num_examples missing in metadata.".format(self.name))
+        try:
+            data = json.loads(model_update.meta)["training_metadata"]
+            num_examples = data["num_examples"]
+        except KeyError as e:
```
ahellander, line 92: I think we should be explicit and print out when "num_examples" are missing (since it is the only mandatory piece of metadata currently. Also, could in theory the keyerror be raised because  "training_metadata" is missing?

ahellander (dismissed):
Looks good in principle. See minor comment. 

I have not tested though.

stefanhellander (approved):

PR #616, opened 2024-05-28
Bug/SK-853 | Helper for package ignored in start_session

Wrede:
start_session helper in the api was  default always numpyhelper, defualt is now "" empty, and will try to fetch helper type from package.

### DIFF
--- a/fedn/network/api/interface.py
+++ b/fedn/network/api/interface.py
@@ -963,7 +963,7 @@ class API:
         round_buffer_size=-1,
         delete_models=True,
         validate=True,
-        helper="numpyhelper",
+        helper="",
         min_clients=1,
         requested_clients=8,
     ):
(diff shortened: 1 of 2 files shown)


fedn/network/api/interface.py
```diff
@@ -1002,13 +1002,17 @@ def start_session(
...
+        if not package:
             return jsonify(
                 {
                     "success": False,
                     "message": "No compute package set. Set compute package before starting session.",
                 }
             )
+        if not helper:
+            # get helper from compute package
+            helper = package["helper"]
```
ahellander, line 1015: Wait, it we can do this here, we should be able to do it also on the combiner? In that case - can we completely eliminate the need to set the helper in the session_config and instead have it encoded in the fedn.yaml?  Even if we in the future will not always have a package, I think we will always have the concept of a Project where a fedn.yaml is used to configure things.

Wrede: I do not completely follow, the package is not used by combiner, sire you could fetch this iformation using the statestore client. Also, this is beyond this PR. This is a small fix.

ahellander (approved):

PR #618, opened 2024-06-05
Bug/SK-875 | Fix health check in combiner

sztoor:
…om 2sec to 20sec. We don't need any extra flags.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -107,7 +107,7 @@ services:
           "/bin/grpc_health_probe",
           "-addr=localhost:12080"
         ]
-      interval: 2s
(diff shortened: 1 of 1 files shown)


sztoor:
All in one combiner fixed.

sztoor:
The "interval" is the gap/pause to start the checks. According to the current settings, the check will start after 20 seconds and will be terminated when the timeout hits. If something fails, the check will start again after 20 seconds. Once it starts, it needs to finish before the timeout; otherwise, it will be terminated again. This process will be repeated for the number of retries set.

Wrede (approved):

PR #619, opened 2024-06-05
Feature/SK-884 | FEDn compatible with Python 3.12

stefanhellander:
## Description

Added Python 3.12 to supported versions, and removed dependency on distutils.


Wrede (changes requested):
you need to add it to the test matrix as well.

Wrede (approved):

PR #638, opened 2024-06-19
Feature/SK-751 | Clients should have unique IDs

niklastheman:
Store client_id as a property in the clients collection. Use this to keep track of clients.


fedn/network/clients/client.py
```diff
@@ -420,6 +422,7 @@ def _listen_to_task_stream(self):
         r = fedn.ClientAvailableMessage()
         r.sender.name = self.name
         r.sender.role = fedn.WORKER
+        r.sender.client_id = self.id
```
ahellander, line 425: Why not call the property 'r.sender.id' ?

Wrede (approved):

PR #641, opened 2024-06-20
Feature/SK-911 | Add terminate session to controller

Wrede:
Terminate an ongoing session when status is "Terminated". Obs, won't flush queues at combiners or terminate ongoing training or validation requests at clients.

requests.patch("http://localhost:8092/api/v1/sessions/b9fa1b07-fef9-4ddf-a869-a1d43f3c9968", json={"status": "Terminated"})

### DIFF
--- a/fedn/network/controller/control.py
+++ b/fedn/network/controller/control.py
@@ -67,6 +67,20 @@ class CombinersNotDoneException(Exception):
         super().__init__(self.message)
 
 
+class SessionTerminatedException(Exception):
+    """Exception class for when session is terminated"""
+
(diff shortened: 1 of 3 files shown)


Wrede:
in progress

ahellander (approved):
Looks like a good start. But I think we will need to think carefully about UX. When terminated on controller, the combiner and clients will still be busy and new sessions will thus be delayed. This won't be the behavior a user expects. I think it would be better if the termination happened at the end of a completed round, and before a new round is started.

niklastheman (approved):

PR #653, opened 2024-07-10
Refactor/SK-936 | Clean up certs and remove /add_combiner logic

Wrede:
This PR cleans up the old cert managment of sending certs via REST from controller to combiner and clients. Certs for gRPC clients can now fully be controlled via environment variables FEDN_GRPC_CERT_PATH which will be read and included as ssl credential in gRPC channels. To control different certs for mutiple combiners one can use FEDN_GRPC_CERT_PATH_"COMBINER NAME" where COMBINER NAME is as stored in the statestore. 
Certs (and the private key) is no longer stored in the statestore (risky), and is not sent over http(s). The server (combiner) can setup certs via "cert_path" and "key_path" which can be provided in the config file. Obs that this is only for gRPC and not REST.

The logic for adding a combiner to the network has changed. Previously, the combiner had to post to the endpoint /add_combiner of the REST API which then returns statestore and modelstore configs (inluding password/access secrets, which is very risky) and added the combiner to statestore. This logic has been removed and instead the combiner requires configs (file) from admin at server start, the combiner will then directly communicate with the statestore server to add the combiner. The combiner in the state does no longer store the certs. However, storage(model) configs is still stored in the statestore (inluding password). This should be removed in a future PR. 

Additional:
- unpinned torch in mnist-pytorch (had problems installing 2.3.1, unsure why)
- Added native config files so that only mongo and minio is required to run via compose. Controller, combiner and clients can run in native mode over localhost network.
- Added typed dict ServerConfig for gRPC server
- setting/downloading the package and saving to file path was hard coded to /app/config (related to docker image) which made it impossibe to use controller in native mode. All /app/config in API has been changed to os.getcwd().
- Client does no longer expect "package" config from controller (if it should be remote or local), the client decide if it want to download a package or use local.
- init of statestore client and modelstore clients on the server side (combiner) has been removed to shared.py for ease of readbility and decrease redudancy.
- Now possible to allow local package using FEDN_ALLOW_LOCAL_PACKAGE env var on controller. If true, clients connecting can use the flag --local-package. TODO: Studio projects should be configurable for this. client are defaulted to numpyhelper (combiner_config will lack helpertype). Should have a flag instead for client to set helper.


Wrede:
Note to self: not good that mongo client starts on import from shared.py

ahellander:
I guess you need to obtain an admin token to post to /add_combiner?

Wrede:
> I guess you need to obtain an admin token to post to /add_combiner?

In what scenario? This is something the user can configure. JWT tokens are not enabled by default.

ahellander:
> > I guess you need to obtain an admin token to post to /add_combiner?
> 
> In what scenario? This is something the user can configure. JWT tokens are not enabled by default.

I was thinking about Studio - it seems like adding a combiner should be the project admin  / API token privilege?

ahellander (approved):
Looks great!

PR #658, opened 2024-07-15
Feature/SK-940 | CLI command for train 

sowmyasris:
Implemented the CLI command for train entrypoint 
Tested the command for mnist-pytorch example 
Needs three arguments 
-p path to fedn.yaml file
-i input model parameter
-o output directory to write the updated model


Wrede (approved):
very good!

PR #666, opened 2024-07-26
Feature/SK-946 | Add functionality for user defined aggregation

viktorvaladi:
Adds initial code for user-defined aggregation.

WIP.


fedn/network/api/client.py
```diff
@@ -574,6 +574,7 @@ def start_session(
         helper: str = "",
         min_clients: int = 1,
         requested_clients: int = 8,
+        function_provider_path: str = None,
```
Wrede, line 577: can't we have the FunctionProvider class as argument instead?

viktorvaladi: That would be better since then the abstract class would prevent wrong structure on the class already at the API level. But I'm not sure what will happened with imports when the class gets stringified.. I'll investigate it :)

Wrede (approved):

PR #667, opened 2024-08-01
Bug/SK-931 | --preferred-combiner should not be boolean

sowmyasris:
Changed --preferred-combiner flag in run_cmd and client_cmd to --combiner
Type change from boolean to string

### DIFF
--- a/fedn/network/clients/connect.py
+++ b/fedn/network/clients/connect.py
@@ -44,16 +44,16 @@ class ConnectorClient:
     :param id: id of client
     """
 
-    def __init__(self, host, port, token, name, remote_package, force_ssl=False, verify=False, grpc_proxy=None,combiner=None, id=None):
+    def __init__(self, host, port, token, name, remote_package, force_ssl=False, verify=False, combiner=None, id=None):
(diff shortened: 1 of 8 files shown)


fedn/network/clients/connect.py
```diff
@@ -38,7 +38,7 @@ class ConnectorClient:
     :param force_ssl: True if https is used, False if http
     :type force_ssl: bool
     :param verify: True if certificate is verified, False if not
-    :type verify: bool
+    :type verify: str
```
Wrede, line 41: why is this changed to "str"?

Wrede (changes requested):
I have changed my mind, let keep the name preferred_combiner otherwise it will have conflicts with new feature --combiner and --proxy-host

fedn/network/clients/connect.py
sowmyasris: When I tried to install using pip3 install -e .(editable mode), it automatically switches to str, even after I changed to bool

Wrede (approved):

PR #674, opened 2024-08-19
Docs/SK-000 | Clarify usage readme

ahellander:
Clarification on free usage.


sztoor (approved):
I think that is what we need, here and on our web pages.

PR #675, opened 2024-08-19
Docs/SK-958 | Improve docs

ahellander:
## Description

This PR imroves the docs based on Salman's experiences from running the tutorial in the workshop. 
- Update Getting Started  to "typical FL flow" to make it more pedagogical (start server, prepare client, start client, run training).  
- Extend "FEDn project" page with more detail and code examples. 
- Updated "Using the Python API" page
- Gathered content related to the local development sandbox under one main section.


sztoor:
I have added a few comments. I think it will be better if we take out the APIClient instructions and make a separate quick stack for API-based communication. The one mentioned in this document will not work until we talk about tokens, URLs etc. I think it will cause confusion.

sztoor (approved):
These changes bring clarity to the instructions.

PR #693, opened 2024-09-03
Feature/SK-933 | Add delete combiner/client to Api

niklastheman:
Added /delete to the combiner/client Api:s.

**Please note:** this only effects the data in mongodb, nothing else.

### DIFF
--- a/fedn/network/api/v1/client_routes.py
+++ b/fedn/network/api/v1/client_routes.py
@@ -368,3 +368,47 @@ def get_client(id: str):
         return jsonify({"message": f"Entity with id: {id} not found"}), 404
     except Exception:
         return jsonify({"message": "An unexpected error occurred"}), 500
+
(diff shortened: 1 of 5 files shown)


ahellander (commented):
Great! 

How does this relate to revoking the token for the client ? Will that client still be able to connect after being deleted (and hence create a new entry in the db ) ?

niklastheman:
> Great!
> 
> How does this relate to revoking the token for the client ? Will that client still be able to connect after being deleted (and hence create a new entry in the db ) ?

Yes exactly. It does not relate to the token what so ever... I just removes the document from the db. You can user either the generated id or client_id for clients.

ahellander (commented):
Looks good to me, but you should fix the codeql warning.

Wrede (approved):

PR #694, opened 2024-09-04
Docs/SK-956 | Update docs to include new CLI entrypoints commands

sowmyasris:
Update the documentation with new CLI entrypoint commands


ahellander (commented):
Great. But are you sure fedn train works this way for the mnist-pytorch example, I wonder about --data-path

sowmyasris:
> Great. But are you sure fedn train works this way for the mnist-pytorch example, I wonder about --data-path

Yes, If there is no --data-path, it picks from
    if data_path is None:
        data_path = os.environ.get("FEDN_DATA_PATH", abs_path + "/data/clients/1/mnist.pt")
         which is specified in data.py.

Wrede:
--data-path is specific for mnist examples. FEDn is not keep track of use-cases data, user have to make sure the training (and test) data is loaded in for whatever command is defined in train entrypoint. So, for mnist examples fedn run startup (to download the data, path will default to "/data/clients/1/mnist.pt" for torch for example ) needs to be run before fedn run train

Wrede:
you can add a note for this in the docs "Please observe that many of the fedn examples requires you to first download the training and test data before you run "fedn run train/validate" entrypoints. The downloads are usually handled by the "fedn run startup" command "

Wrede (approved):

PR #704, opened 2024-09-17
Feature/sk 952

sowmyasris:
Example created for the yolo object detection using darknet


examples/Yolo-object-detection/README.rst
```diff
@@ -0,0 +1,58 @@
+FEDn Project: 'Yolo-object-detection'
```
ahellander, line 1: Can you expand in the readme with more information on what this example is about, maybe include some images etc to explain the ML task being solved, etc.

examples/Yolo-object-detection/client/get_data.py
```diff
@@ -0,0 +1,87 @@
...
+    with open(dir_path + "/data/1.zip", "wb") as f:
+        f.write(response.content)
+    zip_file_path = dir_path + "/data/1.zip"
+    extract_to_path = dir_path + "/data/1/"
+
+    os.makedirs(extract_to_path, exist_ok=True)
+
+    with zipfile.ZipFile(zip_file_path, "r") as zip_ref:
+        zip_ref.extractall(extract_to_path)
+def download_blob():
```
ahellander, line 21: Add space between function definitions.

examples/Yolo-object-detection/client/get_data.py
```diff
@@ -0,0 +1,87 @@
...
+        f.write(response.content)
+    zip_file_path = dir_path + "/data/1/data.zip"
+    extract_to_path = dir_path + "/data"
+
+    os.makedirs(extract_to_path, exist_ok=True)
+
+    with zipfile.ZipFile(zip_file_path, "r") as zip_ref:
+        zip_ref.extractall(extract_to_path)
+
+def train_test_creation(dataset=dir_path + "/data/1/data/", outdir=dir_path + "/data/1/", test_percentage=10):
```
ahellander, line 35: Do we have the capability here to split the dataset into multiple different clients? We do not longer have an environment variable controlling the number of splits?

examples/Yolo-object-detection/client/model.py
```diff
@@ -0,0 +1,39 @@
...
+    else:
+        target_path = parent_dir
+    command = ["git", "clone", repo_url,target_path]
+
+    try:
+        subprocess.run(command, check=True)
+        print(f"Successfully cloned {repo_url}")
+    except subprocess.CalledProcessError as e:
+        print(f"Error during cloning: {e}")
+def init_seed(out_path="../seed.npz"):
```
ahellander, line 24: Add space between function defs.

examples/Yolo-object-detection/client/model.py
```diff
@@ -0,0 +1,39 @@
...
+import numpy as np
+import subprocess
+
+from fedn.utils.helpers.helpers import get_helper
+
+HELPER_MODULE = "numpyhelper"
+helper = get_helper(HELPER_MODULE)
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+def git_clone(repo_url="https://github.com/AlexeyAB/darknet.git", clone_dir="../.darknet"):
```
ahellander, line 11: Why do we need to do a git clone "from code"? Is it not possible to install it as a library in the env?

examples/Yolo-object-detection/client/train.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    os.chdir("../../.darknet")
+    darkfile = dir_path + "/data/1/example.weights"
+    save_fedn2darknet(in_model_path, darkfile)
+    data_file=dir_path + "/data/1/obj.data"
+    cfg_file = dir_path+"/data/1/yolov4-tiny.cfg"
+    darknet_path = "./darknet"  # Make sure this path is correct
+    yolo_converted_weights = dir_path + "/data/1/example.weights"
+    command = [darknet_path, "detector", "train", data_file, cfg_file, yolo_converted_weights,"-dont_show"]
```
ahellander, line 50: Hmm, I see. So you are invoking an extenal command in the end. Would it be possible to skip this wrapping in Python and just use this directly in the entrypoint?

ahellander: This way might work, but now we are running a subprocess in a subprocess...

examples/Yolo-object-detection/requirements.txt
```diff
@@ -0,0 +1,2 @@
+fire==0.3.1
```
ahellander, line 1: I think you are not using any of these?

ahellander (changes requested):

examples/Yolo-object-detection/client/train.py
sztoor: I am not 100% sure, but by a quick search, it seems like Darknet has Python interfaces. It would be neat to use a package rather than a command.

PR #706, opened 2024-09-18
Feature/SK-1039 | Add upload, activate & deactivate package to fedn api v1

niklastheman:
## New endpoints

- **POST** /packages | Uploads a compute package. Stores file in minio and object in mongo. Sets the uploaded package as active.
- **PUT** /packages/active | Expects an "id" property in the body. Sets the provided package (id) as active..
- **DELETE** /packages/active | Deletes the current "active" object (compute package). Thus making no package active. The remove package is still present in the db though, just not in active stage.

### New env

- **FEDN_COMPUTE_PACKAGE_DIR** - config where the compute package is to be stored on disk on the server.

### DIFF
--- a/fedn/network/api/interface.py
+++ b/fedn/network/api/interface.py
@@ -7,7 +7,7 @@ from flask import jsonify, send_from_directory
 from werkzeug.security import safe_join
 from werkzeug.utils import secure_filename
 
-from fedn.common.config import FEDN_ALLOW_LOCAL_PACKAGE, get_controller_config, get_network_config
+from fedn.common.config import FEDN_ALLOW_LOCAL_PACKAGE, FEDN_COMPUTE_PACKAGE_DIR, get_controller_config, get_network_config
 from fedn.common.log_config import logger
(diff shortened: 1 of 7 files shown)


fedn/network/api/interface.py
```diff
@@ -370,22 +370,20 @@ def download_compute_package(self, name):
...
                 with open(file_path, "wb") as fh:
                     fh.write(data)
                 # TODO: make configurable, perhaps in config.py or package.py
-                return send_from_directory(os.getcwd(), name, as_attachment=True)
+                return send_from_directory(FEDN_COMPUTE_PACKAGE_DIR, name, as_attachment=True)
             except Exception:
                 raise
         finally:
-            # Delete the file after it has been saved
-            os.remove(file_path)
```
niklastheman, line 388: @Wrede removed this. Is there a reason you added this or is it a mistake?

Wrede: if "name" is different, and you have many packages, is it not good to delete it?

Wrede (approved):
approved with comment

PR #709, opened 2024-09-22
Feature/SK-948 | Combiner config flags in fedn client 

Wrede:
Using --combiner and --proxy-server the fedn client can now directly connect to the combiner (grpc server) without going via discover service at controller. --combiner is the hostname, proxy_server is e.g grpc.fedn.scaleoutsystems.com


ahellander (approved):
Looks good!

PR #712, opened 2024-09-26
Fix/SK-1060 | Move mnist data to scaleout public bucket

Wrede:
This pull request includes significant changes to the data handling and dependencies in the `examples/mnist-pytorch/client` project. The most important changes involve replacing the use of `torchvision` with direct data downloads, updating the data loading logic, and modifying the environment dependencies.

### Data Handling Improvements:

* [`examples/mnist-pytorch/client/data.py`](diffhunk://#diff-8c45613695cb7a03b1da1922cc8394aa271841a45fdc0922fc1d601ab5784712R4-R32): Replaced `torchvision` dataset download with direct data download using `requests`. This includes setting a random split ID and downloading the corresponding data file from Scaleout's own public bucket.
* [`examples/mnist-pytorch/client/data.py`](diffhunk://#diff-8c45613695cb7a03b1da1922cc8394aa271841a45fdc0922fc1d601ab5784712L34-R58): Updated the `load_data` function to dynamically find the correct data split ID and verify the existence of the data file.
* [`examples/mnist-pytorch/client/data.py`](diffhunk://#diff-8c45613695cb7a03b1da1922cc8394aa271841a45fdc0922fc1d601ab5784712L51-L97): Removed the `splitset` and `split` functions, as the data splitting is now handled by the pre-downloaded data files.

### Dependency Updates:

* [`examples/mnist-pytorch/client/python_env.yaml`](diffhunk://#diff-1e30a56a7e6d471ce83c548eddd8c01c9b6d29c37088701b92d3b0f65111463aL11-L12): Removed `torchvision` from the dependencies and updated `numpy` versions to ensure compatibility with different platforms and Python versions.

### DIFF
--- a/examples/mnist-pytorch/client/data.py
+++ b/examples/mnist-pytorch/client/data.py
@@ -1,5 +1,4 @@
 import os
-from math import floor
 
 import numpy as np
 import requests
--- a/examples/mnist-pytorch/client/data.py
+++ b/examples/mnist-pytorch/client/data.py
@@ -14,7 +14,7 @@ def get_data(out_dir="data"):
 
     if not os.path.exists(f"{abs_path}/{out_dir}/clients/{split_id}"):
         # create directory for data
-        os.makedirs(f"{abs_path}/{out_dir}data/clients/{split_id}")
+        os.makedirs(f"{abs_path}/{out_dir}/clients/{split_id}")
 
     # use requests to download the data from url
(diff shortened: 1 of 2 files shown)


ahellander (changes requested):
I think this is a good general improvement, although it reduces the flexibility a bit (max 10 splits possible).  

But you need to update the getting started guide as well, e.g. remove the FEDN_NUM_DATA_SPLITS env variable. 

Not needed now ,but I think for pedagogical reasons it would also be good to add the option to first download the data split,  then start the client. But this could be done later.

PR #722, opened 2024-10-07
Feature/SK-1071 | Initial refactor of client

niklastheman:
### fedn client start-v2 command added

OBS this PR does not replace the current client, it's and initial step to a new client (refactor)

Mainly tried to clean up the existing client. The existing client is left as is. The idea is to let this new solution exist in parallel until it can be the main solution. The functionality of the client is divided into three main classes, GrpcHandler, ClientApi and the "new" Client class. The Client class implements the ClientApi. The Client class can be understood as an example of how to set up your own client.

### DIFF
--- a/fedn/network/clients/client_api.py
+++ b/fedn/network/clients/client_api.py
@@ -1,6 +1,5 @@
 import enum
 import os
-import threading
 import time
 from io import BytesIO
 from typing import Any, Tuple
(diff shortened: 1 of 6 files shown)


Wrede (approved):
I want @ahellander to look at this also

fedn/network/clients/client_api.py
```diff
@@ -0,0 +1,321 @@
...
+        self.package = package
+        self.ip = ip
+        self.port = port
+        self.helper_type = helper_type
+
+
+# Enum for respresenting the result of connecting to the FEDn API
+class ConnectToApiResult(enum.Enum):
+    Assigned = 0
+    ComputePackgeMissing = 1
```
ahellander, line 31: Spelling error .

fedn/network/clients/client_api.py
```diff
@@ -0,0 +1,321 @@
...
+        self.grpc_handler: GrpcHandler = None
+
+    def subscribe(self, event_type: str, callback: callable):
+        """Subscribe to a specific event."""
+        if event_type in self._subscribers:
+            self._subscribers[event_type].append(callback)
+        else:
+            raise ValueError(f"Unsupported event type: {event_type}")
+
+    def notify_subscribers(self, event_type: str, *args, **kwargs):
```
ahellander, line 69: I don't understand, what subscribers might a client have?

ahellander (approved):
Looks like a good refactor, I added two minor comments, but I'm sure this will be iterated on a number of times before replacing v1.

fedn/network/clients/client_api.py
niklastheman: @ahellander if you look in the client_v2.py file you can see how it is used. The overall idea is that client_api exposes functionality that a client can then make use of... In this particular example the client subscribes to events raised from the client_api. Events include training, validation or inference.

niklastheman: I will add a little example too.

fedn/network/clients/client_api.py
niklastheman: Will fix this.

PR #726, opened 2024-10-11
Feature/SK-1102 | Simplify ClientAPI and add support for inference

niklastheman:
Added a readme file to the client directory. This is an attempt to make users see how easy they can create their own clients.

Benjamin:
**Key changes**
* Moved boiler plate code from client to ClientAPI to make it simpler for the user to implement an on_train/on_validate function
* Updated client_v2 accordingly
* These changes **eliminate the client's** (client_v2 or exemple in README) **need to import fedn protos**
* Added support for _predict_ in ClientAPI and GrpcHandler
* Added functions in GrpcHandler to create gRPC messages for model update, validation and prediction

grpc_handler.py
* Added function `create_update_message`, `create_validation_message` and `create_prediction_message` which populate and return a gRPC message of the relevant type
* Function `send_model_update()` now takes a ModelUpdate message as input directly and only handles logic for sending the message to the server

client_api.py
* Replaced `_subscribers` with `train_callback` and `validate_callback`
  - `train_callback` needs to take in_model as parameter and return out_model and metadata dict
  - `validate_callback` needs to take in_model as parameter and return metrics dict
* `_task_stream_callback` now invokes `update_local_model` or `validate_global_model` depending on task request type
  - `update_local_model` and `validate_global_model` contain boiler plate code that was previously inside the `on_train` and `on_validate` callbacks (getting/sending model from/to combiner, generating uuid, assembling metadata, sending model update/validation)
  - This replaces the functions `train` and `validate`
* `run` method creates a child thread for heart beats and listens to task stream in main thread
* gRPC messages for update, validation and predict are created in `update_local_model`, `validate_global_model` and `predict_global_model` respectively and are sent in status and to GrpcHandler

client_v2.py
* Adjusted training and validation functions for the updated client_api.py

README.rst
* Updated example client that makes use of the new API


ahellander (commented):
Nice, I think it will be helpful to have something like this. 

However I think we can make it even neater, I have some ideas, let's talk about it.

niklastheman:
> Nice, I think it will be helpful to have something like this.
> 
> However I think we can make it even neater, I have some ideas, let's talk about it.

Yes, let's!

fedn/network/clients/README.rst
```diff
@@ -0,0 +1,126 @@
...
+            "round_id": 1,
+        }
+
+        metadata = {
+            "training_metadata": training_metadata,
+            "config": json.dumps(config),
+        }
+
+        # Do your training here, out_model is your result...
+        out_model = in_model
```
ahellander, line 58: I think it is better to give a minimal, but working example. This could be confusing.

fedn/network/clients/README.rst
```diff
@@ -0,0 +1,126 @@
...
+
+    def on_train(in_model):
+        training_metadata = {
+            "num_examples": 1,
+            "batch_size": 1,
+            "epochs": 1,
+            "lr": 1,
+        }
+
+        config = {
```
ahellander, line 48: Don't understand this. How can the client set the round_id?

fedn/network/clients/README.rst
```diff
@@ -0,0 +1,126 @@
...
+        metrics = {
+            "test_accuracy": 0.9,
+            "test_loss": 0.1,
+            "train_accuracy": 0.8,
+            "train_loss": 0.2,
+        }
+        return metrics
+
+    def main(api_url: str, api_port: int, token: str = None):
+        print(f"API URL: {api_url}")
```
ahellander, line 74: Remove printouts?

benjaminastrand:
- Added predict to GrpcHandler and ClientAPI
- Added functions to create gRPC messages for update, validation and prediction
   * ClientAPI send request + model data to GrpcHandler and receives gRPC message back
   * gRPC message is sent to combiner and in status
- Updated README (minimal example of ML training will be added in a new issue)

Wrede (commented):
Imo I think you should change the Name of ClientAPI to maybe Project, ProjectManager, Context, ContextManager and import it into the root __init__ so that you can:
from fedn import Context

benjaminastrand:
Changed name of client api
- ClientAPI -> FednClient
- client_api.py -> fedn_client.py

Wrede (approved):

PR #731, opened 2024-10-29
Refactor/SK-1144 | Simplify on_train and on_validate

benjaminastrand:
**Key changes**
* Moved boiler plate code from client to ClientAPI to make it simpler for the user to implement an on_train/on_validate function
* Updated client_v2 accordingly
* These changes **eliminate the client's** (client_v2 or exemple in README) **need to import fedn protos**

grpc_handler.py
* Removed unnecessary params from `send_model_update()`

client_api.py
* Replaced `_subscribers` with `train_callback` and `validate_callback`
  - `train_callback` needs to take in_model as parameter and return out_model and metadata dict
  - `validate_callback` needs to take in_model as parameter and return metrics dict
* `_task_stream_callback` now invokes `update_local_model` or `validate_global_model` depending on task request type
  - `update_local_model` and `validate_global_model` contain boiler plate code that was previously inside the `on_train` and `on_validate` callbacks (getting/sending model from/to combiner, generating uuid, assembling metadata, sending model update/validation)
  - This replaces the functions `train` and `validate`
* `run` method creates a child thread for heart beats and listens to task stream in main thread

client_v2.py
* Adjusted training and validation functions for the updated client_api.py

README.rst
* Updated example client that makes use of the new API

### DIFF
--- a/fedn/network/clients/README.rst
+++ b/fedn/network/clients/README.rst
@@ -75,7 +75,7 @@ Step-by-Step Instructions
         print(f"API Token: {token or "-"}")
         print(f"API Port: {api_port or "-"}")
 
-        client_api = ClientAPI()
+        client_api = ClientAPI(train_callback=on_train, validate_callback=on_validate)
 
         url = get_api_url(api_url, api_port)
 
@@ -102,9 +102,6 @@ Step-by-Step Instructions
 
         if not result:
             return
-
-        client_api.set_train_callback(on_train)
-        client_api.set_validate_callback(on_validate)
 
         client_api.run()
 
--- a/fedn/network/clients/README.rst
+++ b/fedn/network/clients/README.rst
@@ -26,110 +26,97 @@ Step-by-Step Instructions
 
     import argparse
     import json
-    import threading
     import uuid
-    
-    import fedn.network.grpc.fedn_pb2 as fedn
+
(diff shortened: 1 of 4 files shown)


fedn/network/clients/README.rst
niklastheman, line 1: I think we could include a callback function for training in this README example. It doesn't need to do anything more than return the input model. The end goal is to make users understand how they can develop clients that help with their use case.

fedn/network/clients/client_api.py
```diff
@@ -230,36 +336,34 @@ def send_status(self, msg: str, log_level=fedn.Status.INFO, type=None, request=N
...
-            sender_name=sender_name,
-            sender_role=sender_role,
-            client_id=client_id,
+            sender_name=self.name,
             model_id=model_id,
             model_update_id=model_update_id,
-            receiver_name=receiver_name,
-            receiver_role=receiver_role,
+            receiver_name=request.sender.name,
+            receiver_role=request.sender.role,
```
niklastheman, line 350: Not sure about these changes... Don't know if a TaskRequest as input is the way to go. It creates a dependency that did not exist previously... If you think it's cleaner, keep it.

niklastheman (approved):
Good improvements overall, nice work! I have some input (nothing major), check out the comments!

PR #738, opened 2024-11-05
Docs/SK-1173 | Quickstart guide updated with the Packages link in the studio

sztoor:
## Description

I have updated the quickstart guide with the Packages link in the studio. Also, I have highlighted that there is a possibility to select different packages and seed models.


Wrede (changes requested):
upload_seed.png looks like compute package upload? 

I would I to the note about using local package: "This is a good option for when developing your own project so that you don't have to upload a new package on every change/version"

sztoor:
I have added more information about the local compute package and uploaded a new upload_seed.png file.

Wrede (commented):
Will it not look strange having images that are "zoom in"/font size differs, comparing to other images on "getting started"?

PR #739, opened 2024-11-05
Fix/SK-1178 | Set client_id in ModelRequest

Wrede:
Use client_id for the client id instead of setting "name" to be the id.

### DIFF
--- a/fedn/network/clients/grpc_handler.py
+++ b/fedn/network/clients/grpc_handler.py
@@ -17,6 +17,24 @@ from fedn.common.config import FEDN_AUTH_SCHEME
 from fedn.common.log_config import logger
 from fedn.network.combiner.modelservice import upload_request_generator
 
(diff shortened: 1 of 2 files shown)


fedn/network/clients/grpc_handler.py
```diff
@@ -191,7 +218,7 @@ def get_model_from_combiner(self, id: str, client_name: str, timeout: int = 20)
         data = BytesIO()
         time_start = time.time()
         request = fedn.ModelRequest(id=id)
-        request.sender.name = client_name
+        request.sender.client_id = client_name
```
benjaminastrand, line 221: Parameter `client_name` in `get_model_from_combiner` is the client name - not the client id - as this function is currently used.

Maybe the parameter `client_name` should be replaced by `client_id`?

niklastheman (approved):

PR #741, opened 2024-11-06
Fix/SK-1181 | Prevent default flag values from always overwriting yaml file

benjaminastrand:
- Previously default values for name, trainer and validator were set using click - which meant they always overwrote the values in yaml file.
- Logic for setting default values is now moved to function body.
- Now, the logic works as presented in the table below. The entries show which value is used in each of the 4 situations.

| Flag\File    | Set | Not set |
| -------- | ------- | ------- |
| Set  | Flag value    | Flag value    |
| Not set | File value    | Default value   |


Wrede (approved):
I like the table, we should also add env_vars

PR #754, opened 2024-11-18
Feature/SK-1217 | Add client settings to client v2

viktorvaladi:
Adds client settings to client v2 so that the UDF code is functional.


Wrede (approved):
Add @benjaminastrand and @niklastheman as reviewer also I want 1 approval from them.

benjaminastrand (changes requested):
Since client_options is saved inside `_process_training_request` in client_v2.py, it will only be saved if the user chooses to start the client through `fedn client start`.

I suggest saving the metadata in `update_local_model` in fedn_client.py instead.

viktorvaladi:
> Since client_options is saved inside `_process_training_request` in client_v2.py, it will only be saved if the user chooses to start the client through `fedn client start`.
> 
> I suggest saving the client_options in `update_local_model` in fedn_client.py instead.

As discussed on discord, its good to have it as it is in the PR? So for the on_train() function it would be the client_settings parameter added which defaults to an empty dict if its not used?

benjaminastrand (approved):

PR #758, opened 2024-11-22
Refactor/SK-1225 | Use api/v1 in APIClient start_session

benjaminastrand:
The function `start_session` now makes two POST requests in sequence instead of one:
1. `host_url/api/v1/sessions` - to create a session
2. `host_url/api/v1/sessions/start` - to start the session

The 2nd request is made only if the 1st request returns status code 201, indicating that the session has been created, and the response from the 2nd request is returned. Otherwise, the response from the 1st request is returned.

Also, the method `set_active_model` in APIClient now sets the active helper based on the file extension of the seed model file used (".npz" -> numpyhelper, ".bin" -> binaryhelper)

### DIFF
--- a/fedn/network/api/client.py
+++ b/fedn/network/api/client.py
@@ -645,6 +645,9 @@ class APIClient:
             headers=self.headers,
         )
 
+        if id is None:
+            id = response.json()["session_id"]
+
         if response.status_code == 201:
             response = requests.post(
                 self._get_url_api_v1("sessions/start"),
--- a/fedn/network/api/client.py
+++ b/fedn/network/api/client.py
@@ -607,25 +607,38 @@ class APIClient:
         :rtype: dict
         """
         response = requests.post(
-            self._get_url("start_session"),
+            self._get_url_api_v1("sessions"),
             json={
                 "session_id": id,
-                "aggregator": aggregator,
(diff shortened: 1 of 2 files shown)


fedn/network/api/client.py
```diff
@@ -607,26 +607,39 @@ def start_session(
...
+                    "validate": validate,
+                    "helper_type": helper,
+                    "server_functions": None if server_functions is None else inspect.getsource(server_functions),
+                },
             },
             verify=self.verify,
             headers=self.headers,
         )
 
+        if response.status_code == 201:
```
niklastheman, line 631: Cool! An improvement could be to indicate to the user what has gone wrong if status code != 201.

niklastheman (approved):
Nice work! I left a comment.

benjaminastrand:
Setting the active helper in `set_active_model` so that the user doesn't need to upload a package to set the helper type via APIClient.

niklastheman (approved):

fedn/network/api/client.py
benjaminastrand: The method will return the result from the first request, so the user already has access to this information.

niklastheman (approved):

PR #759, opened 2024-11-25
Feature/SK-1202 | Added log-in functionality

KatHellg:
Added log-in functionality by saving access token to a context file.


fedn/cli/login_cmd.py
```diff
@@ -0,0 +1,58 @@
+import yaml
+import click
+import requests
+
+from .main import main
+from getpass import getpass
+
+# Replace this with the platform's actual login endpoint
+URL = "https://fedn.scaleoutsystems.com/api/token/"
```
niklastheman, line 9: **Cool stuff!** This one should be configurable by the user. Perhaps default to _https://fedn.scaleoutsystems.com_ (and then concat the url programatically). This is because FEDn Studio can be hosted on different domains and we would like this to work on all instances.

fedn/cli/client_cmd.py
```diff
@@ -60,17 +60,47 @@ def list_clients(ctx, protocol: str, host: str, port: str, token: str = None, n_
...
-    click.echo(f"Headers: {headers}")
 
     try:
         response = requests.get(url, headers=headers)
-        print_response(response, "clients")
+        print_response(response, "clients", None)
     except requests.exceptions.ConnectionError:
         click.echo(f"Error: Could not connect to {url}")
 
+@click.option("-p", "--protocol", required=False, default=CONTROLLER_DEFAULTS["protocol"], help="Communication protocol of controller (api)")
```
niklastheman, line 70: Why are these changes in this pr?

niklastheman (changes requested):

Wrede (approved):

niklastheman (approved):

PR #760, opened 2024-11-27
Update apiclient.rst

urwithajit9:
This is to make clear about api endpoint.


Wrede:
Thanks! However, we are currently working already to improve this part of the docs in a greater context.

PR #762, opened 2024-11-29
Fix/SK-1234 | Remove "self" in self.logger

benjaminastrand:
Remove "self" in self.logger(... in `grpc_handler`


viktorvaladi (approved):
LGTM

PR #792, opened 2025-01-27
Bugfix/SK-1253 | Client exit if no python_env is defined in fedn.yaml

Wrede:
If python_env is removed from fedn.yaml the client will exit and assume there should always be an env. This is not the expected behavior. python_env is an optional feature. 

This PR fixes that.


ahellander (approved):

niklastheman (approved):
Perhaps comment should change if not successfull?

PR #793, opened 2025-01-27
Feature/SK-1330 | Use host cert store instead of downloading cert

stefanhellander:
## Description

Removes download of remote cert, use host cert store instead.


Wrede:
This was causing issues before, you mean we don't need it anymore?

Wrede (approved):

Issue #234, opened 2021-06-22
Multiple native clients not compatible with current log UI assuming port 8080 is allocatable.


morganekmefjord:
**Severity**
- [ ] Critical/Blocker (select if the issue makes the application unusable or causes serious data loss)
- [ ] High (select if the issue affects a major feature and there is no workaround or the available workaround is very complex)
- [x] Medium (select if the issue affects a minor feature or affects a major feature but has an easy enough workaround to not cause any major inconvenience)
- [] Low (select if the issue doesn't significantly affect the user experience, like minor visual bugs)

**Describe the bug**
When starting multiple clients on the same host machine running natively without config the default port is (obviously) in use for serving the client log UI. 

**Suggested mitigation**
 - Reverse the logic and disable the local UI by default and allow for parametric enabling of the local logging UI
 - We should update document about that also that exposing that endpoint you can relay also to any logging services your org might be using for aggregation of infra/service logs
 - Make configuration of port easier. 

**Furthermore**
- Also should we recover and try another port?

ahellander:
I think the UI should be disabled by default (since we are not yet using it).

ahellander:
Has been disabled now (but code is still there). Closing this, but we should later make a CLI flab to toggle it on/off.

PR #226, opened 2021-06-19
File exist check option

morganekmefjord:
## Status

- [x] Ready
- [ ] Draft
- [ ] Hold

## Description
Added functionality to check if files exist in the client side and warn if overwrite parameter is set to false.
Not sure this is useful if we _by convention_ imply that client/ dir must be ok to overwrite. And if using local files --remote=false should be set. I see cornercases though when this functionality is useful.

## Types of changes

What types of changes does your code introduce to FEDn?

- [ ] Hotfix (fixing a critical bug in production)
- [ ] Bugfix
- [x] New feature
- [ ] Documentation update

## Checklist

_If you're unsure about any of the items below, don't hesitate to ask. We're here to help! 
This is simply a reminder of what we are going to look for before merging your code._

- [x] This pull request is against **develop** branch (not applicable for hotfixes)
- [ ] I have included a link to the issue on GitHub or JIRA (if any)
- [ ] I have included migration files (if there are changes to the model classes)
- [ ] I have read the [CONTRIBUTING](https://github.com/scaleoutsystems/fedn/blob/master/CONTRIBUTING.md) doc
- [ ] I have included tests to complement my changes
- [ ] I have updated the related documentation (if necessary) 
- [x] I have added a reviewer for this pull request
- [x] I have added myself as an author for this pull request

## Further comments

Anything else you think we should know before merging your code!

PR #794, opened 2025-01-27
Fix/SK-1320 | Fix vuln.  #824, #234, #230, #226, #216 (perl & kerberos)

Wrede:
Ongoing work.


Wrede:
closing, will work on a more lightweight image like alpine instead.

PR #795, opened 2025-01-27
Feature/SK-1339 | Add predictions (list etc) to APIClient

niklastheman:
### Add predictions (list etc) to APIClient

**Changes:**
- mnist-pytorch predict example updated to store json
- name **route change**:  predict => predictions **!OBS**
- list predictions added to api client
- start predictions added to api client


benjaminastrand (approved):
Looks good!

PR #800, opened 2025-01-28
Feature/SK-1340 | Add stacked bar chart graph for predictions per client

niklastheman:
### Added stacked bar chart

**See image:**

Screenshot 2025-01-28 at 13 47 24


benjaminastrand (changes requested):
Looks very good!
* I would replace "digit" with "class" or "predicted class" in the graph
* In the future, we should use an example where we do inference on a more unbalanced dataset so that the stacks in the graph are not all the same size (but I don't think this is a priority in this stage)

benjaminastrand (approved):

PR #804, opened 2025-01-29
Feature/SK-1298 | Log unexpected exceptions in routes

niklastheman:
Added logger to log exceptions in routes

### DIFF
--- a/fedn/network/api/v1/client_routes.py
+++ b/fedn/network/api/v1/client_routes.py
@@ -1,6 +1,7 @@
 from flask import Blueprint, jsonify, request
 
 from fedn.common.config import get_controller_config, get_network_config
+from fedn.common.log_config import logger
 from fedn.network.api.auth import jwt_auth_required
 from fedn.network.api.shared import client_store, control, get_checksum, package_store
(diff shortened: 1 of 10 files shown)


carl-andersson (commented):
Det är bara okända exceptions som loggas antar jag? model_routes.py har någon loggning som inte loggas på samma sätt, by design?  package_routes.py ->  upload_package loggar inte alla unknown exception.

niklastheman:
> Det är bara okända exceptions som loggas antar jag? model_routes.py har någon loggning som inte loggas på samma sätt, by design? package_routes.py -> upload_package loggar inte alla unknown exception.

Yes, unknown exceptions only. There is only one exception raised from the stores and that is the "EntityNotFound" exception. Some of these (get methods) should be removed. 

Cannot find the difference between model_routes and others...

Nice catch on the upload package endpoint, it has been updated!

carl-andersson:
In model_routes there is this example with nested try catch 
```
except Exception as e:
    logger.error("Error occured during model loading")
    logger.debug(e)
```

niklastheman:
> In model_routes there is this example with nested try catch
> 
> ```
> except Exception as e:
>     logger.error("Error occured during model loading")
>     logger.debug(e)
> ```

Cool man, this has been replaced.

PR #807, opened 2025-01-31
Feature/SK-1367 | Refactor the creation of the databaseconnection

carl-andersson:
This pull request introduce changes to the `fedn` network API and storage system. The primary focus is on refactoring the database connection setup and store access to use a new `DatabaseConnection` class, which centralizes and simplifies the configuration and initialization of database connections and primarily enables testing.

### DIFF
--- a/fedn/network/api/shared.py
+++ b/fedn/network/api/shared.py
@@ -40,12 +40,12 @@ repository = Repository(modelstorage_config["storage_config"])
 
 control = Control(
     network_id=network_id,
-    session_store=stores.session_store,
-    model_store=stores.model_store,
-    round_store=stores.round_store,
-    package_store=stores.package_store,
-    combiner_store=stores.combiner_store,
-    client_store=stores.client_store,
+    session_store=session_store,
+    model_store=model_store,
+    round_store=round_store,
+    package_store=package_store,
+    combiner_store=combiner_store,
+    client_store=client_store,
     model_repository=repository,
 )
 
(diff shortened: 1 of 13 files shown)


fedn/network/api/shared.py
```diff
@@ -1,82 +1,51 @@
...
 
 control = Control(
     network_id=network_id,
-    session_store=session_store,
-    model_store=model_store,
-    round_store=round_store,
-    package_store=package_store,
-    combiner_store=combiner_store,
-    client_store=client_store,
+    session_store=stores.session_store,
```
niklastheman, line 43: Can these be changed from: stores.session_store => session_store? etc...

fedn/network/api/shared.py
```diff
@@ -94,7 +63,7 @@ def get_checksum(name: str = None):
 
     if name is None:
         try:
-            active_package = package_store.get_active()
+            active_package = DatabaseConnection().get_stores().package_store.get_active()
```
niklastheman, line 66: Can this be changed to package_store.get_active()?

niklastheman (commented):
It looks brilliant, I only have a couple of questions.

fedn/network/api/shared.py
carl-andersson: Consider it done, even though I'd rather want to create the Control object as singleton later also

fedn/network/api/shared.py
carl-andersson: I changed it now, however, I do believe that this function does not belong here and if you move it I believe this implemention is to prefer

niklastheman (approved):

PR #808, opened 2025-02-06
Feature/SK-1289 | Added test for store list functionality

carl-andersson:
This pull request includes various changes to improve the testing setup, database connection handling. The most important changes include adding new VS Code configurations for testing, updating database connection methods.

### DIFF
--- a/fedn/network/storage/statestore/stores/client_store.py
+++ b/fedn/network/storage/statestore/stores/client_store.py
@@ -238,8 +238,10 @@ class SQLClientStore(ClientStore, SQLStore[Client]):
 
                 stmt = stmt.order_by(sort_obj)
 
-            if limit != 0:
+            if limit:
                 stmt = stmt.offset(skip or 0).limit(limit)
+            elif skip:
+                stmt = stmt.offset(skip)
 
             items = session.scalars(stmt).all()
 
--- a/fedn/network/storage/statestore/stores/prediction_store.py
+++ b/fedn/network/storage/statestore/stores/prediction_store.py
@@ -123,10 +123,10 @@ class SQLPredictionStore(PredictionStore, SQLStore[Prediction]):
                 correlation_id=item.get("correlationId") or item.get("correlation_id"),
                 data=item.get("data"),
                 model_id=item.get("modelId") or item.get("model_id"),
-                receiver_name=receiver.get("name"),
-                receiver_role=receiver.get("role"),
-                sender_name=sender.get("name"),
-                sender_role=sender.get("role"),
+                receiver_name=receiver.get("name") if receiver else None,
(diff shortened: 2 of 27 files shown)


fedn/network/storage/statestore/stores/prediction_store.py
```diff
@@ -123,10 +123,10 @@ def add(self, item: Prediction) -> Tuple[bool, Any]:
                 correlation_id=item.get("correlationId") or item.get("correlation_id"),
                 data=item.get("data"),
                 model_id=item.get("modelId") or item.get("model_id"),
-                receiver_name=receiver.get("name"),
-                receiver_role=receiver.get("role"),
-                sender_name=sender.get("name"),
-                sender_role=sender.get("role"),
+                receiver_name=receiver.get("name") if receiver else None,
```
niklastheman, line 126: The get method returns None if not present no? At least I think you can set a default: get("name", None)... But I might be misstaken...

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -238,8 +238,10 @@ def list(self, limit: int, skip: int, sort_key: str, sort_order=pymongo.DESCENDI
 
                 stmt = stmt.order_by(sort_obj)
 
-            if limit != 0:
+            if limit:
```
niklastheman, line 241: We could perhaps check so that these values are positive... (This and similar)

fedn/tests/stores/test_client_store.py
niklastheman, line 1: Sweet but can this not be done with a class based approach? Similar to how we have the store set up I imagine... The stores should work similar across different entities so there should be some common ground. It's a thought but I think I could help when adding when adding tests for updating, deleting etc...

fedn/network/storage/statestore/stores/client_store.py
carl-andersson: Yeah, it might be a good idea. For now I was happy with just supporting the current usage. It seemed that both None and 0 was used interchangably there so I figure that the code should support that

fedn/network/storage/statestore/stores/prediction_store.py
carl-andersson: If reciever is None you can't call .get() on it. That's the problem

niklastheman: Ahh yeah, my misstake

fedn/network/storage/statestore/stores/client_store.py
niklastheman: This too is due to how the stores have been used (only via the routes). Previously we have programatically ensured limit will never be null... This is an improvement as is, do with the comment what you will.

PR #809, opened 2025-02-07
Feature/SK-1343 | Enable use of any S3 backend

stefanhellander:
## Description

 Enables use of any S3 backend also when deployed via Studio.

### DIFF
--- a/fedn/network/api/shared.py
+++ b/fedn/network/api/shared.py
@@ -56,9 +56,10 @@ control = Control(
 # TODO: use Repository
 minio_repository: RepositoryBase = None
 
-if modelstorage_config["storage_type"] == "S3":
+storage_type = os.environ.get("FEDN_STORAGE_TYPE", modelstorage_config["storage_type"])
+if storage_type == "S3":
     minio_repository = MINIORepository(modelstorage_config["storage_config"])
-elif modelstorage_config["storage_type"] == "SAAS":
+elif storage_type == "SAAS":
     minio_repository = SAASRepository(modelstorage_config["storage_config"])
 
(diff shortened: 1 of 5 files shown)


niklastheman (approved):
Might be outside the scope of this pr but we could reflect about whether or not the Repository class is needed. Controller etc, could accept **RepositoryBase** as input which in turn then would be realized by **MinioRepository** or **SaasRepositoty**. There might be an explanation that for me is unclear, but it's an observation :)

stefanhellander:
> Might be outside the scope of this pr but we could reflect about whether or not the Repository class is needed. Controller etc, could accept **RepositoryBase** as input which in turn then would be realized by **MinioRepository** or **SaasRepositoty**. There might be an explanation that for me is unclear, but it's an observation :)

Absolutely, I think this should be cleaned up more. Probably should get rid of miniorepository.py as well and have one S3 class, but want to keep these changes as minimal as possible for now to ensure this also works with docker compose as well as old projects in the SaaS for now.

niklastheman:
> > Might be outside the scope of this pr but we could reflect about whether or not the Repository class is needed. Controller etc, could accept **RepositoryBase** as input which in turn then would be realized by **MinioRepository** or **SaasRepositoty**. There might be an explanation that for me is unclear, but it's an observation :)
> 
> Absolutely, I think this should be cleaned up more. Probably should get rid of miniorepository.py as well and have one S3 class, but want to keep these changes as minimal as possible for now to ensure this also works with docker compose as well as old projects in the SaaS for now.

Sounds good!

Wrede (approved):
GJ!

PR #814, opened 2025-02-18
Feature/SK-1269 | EntityNotFoundException removed from State Stores

FrankJonasmoelle:
## EntityNotFoundException removed from State Stores

This PR changes how exceptions are handled in the state stores and model routes.

### Before:
- In the state stores, an exception was raised when an entity was not found (EntityNotFoundException).
- The api routes used an try except statement to catch the exceptions and returned the error message.

### Now:
- If an entity is not found, the state stores return None.
- The api routes check whether an entity is None and return an error message in that case.


fedn/network/api/v1/model_routes.py
```diff
@@ -511,11 +515,12 @@ def get_descendants(id: str):
 
         descendants = model_store.list_descendants(id, limit or 10)
 
-        response = descendants
+        if descendants is None:
```
niklastheman, line 518: Will this be None if initial model is not found? Empty list is fine to return if there are no descendants of the specified model.

fedn/network/api/v1/model_routes.py
```diff
@@ -580,12 +585,10 @@ def get_ancestors(id: str):
         include_self: bool = include_self_param and include_self_param.lower() == "true"
 
         ancestors = model_store.list_ancestors(id, limit or 10, include_self=include_self, reverse=reverse)
-
+        if ancestors is None:
```
niklastheman, line 588: Similar question as descendants :) (Please note that this might absolutely be correct, it's simply a question)

fedn/network/api/v1/session_routes.py
```diff
@@ -458,14 +458,14 @@ def patch_session(id: str):
             session[key] = value
 
         success, message = session_store.update(_id, session)
+        if success is False:
+            jsonify({"message": f"Failed to update session: {message}"}), 500
```
niklastheman, line 462: I think 500 is incorrect to return here... Please check but think 400 is more appropriate... 500 is for unhandled exceptions...

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -190,7 +190,7 @@ def update(self, id: str, item: Client) -> Tuple[bool, Any]:
             existing_item = session.scalars(stmt).first()
 
             if existing_item is None:
-                raise EntityNotFound(f"Entity with (id | client_id) {id} not found")
+                return False, "Item not found"
```
niklastheman, line 193: The update methods are possibly the one place we could raise the exception... I don't know what is best, I think both solutions work. We could perhaps take 2 min and talk it through...

fedn/network/storage/statestore/stores/model_store.py
```diff
@@ -137,7 +137,7 @@ def list_descendants(self, id: str, limit: int) -> List[Model]:
         model: object = self.database[self.collection].find_one(kwargs)
 
         if model is None:
-            raise EntityNotFound(f"Entity with (id | model) {id} not found")
+            return None
```
niklastheman, line 140: This looks good, think it answers question above...

PR #817, opened 2025-02-19
Feature/SK-1270 | Change interface of ClientStore to use the Client class as data container

carl-andersson:
The changes introduce a new `Client` datamodel and updates methods to use this datamodel

### DIFF
--- a/fedn/network/api/network.py
+++ b/fedn/network/api/network.py
@@ -85,6 +85,7 @@ class Network:
         # TODO: Implement strategy to handle an unavailable combiner.
         logger.warning("REDUCER CONTROL: Combiner {} unavailable.".format(combiner.name))
 
+    # TODO: This method is not used in the current version of FEDn.
     def add_client(self, client):
         """Add a new client to the network.
 
@@ -98,6 +99,7 @@ class Network:
         logger.info("adding client {}".format(client["client_id"]))
         self.client_store.upsert(client)
 
+    # TODO: This method is not used in the current version of FEDn.
     def get_client(self, client_id: str):
         """Get client by client_id.
 
@@ -112,6 +114,7 @@ class Network:
         except Exception:
             return None
 
+    # TODO: This method is not used in the current version of FEDn.
     def get_client_info(self):
         """List available client in statestore.
 
--- a/fedn/network/api/network.py
+++ b/fedn/network/api/network.py
@@ -5,6 +5,7 @@ from fedn.network.combiner.interfaces import CombinerInterface
 from fedn.network.loadbalancer.leastpacked import LeastPacked
 from fedn.network.storage.statestore.stores.client_store import ClientStore
 from fedn.network.storage.statestore.stores.combiner_store import CombinerStore
+from fedn.network.storage.statestore.stores.dto import Client
 
 __all__ = ("Network",)
 
@@ -86,7 +87,7 @@ class Network:
         logger.warning("REDUCER CONTROL: Combiner {} unavailable.".format(combiner.name))
 
     # TODO: This method is not used in the current version of FEDn.
-    def add_client(self, client):
+    def add_client(self, client: Client):
         """Add a new client to the network.
 
         :param client: The client instance object
@@ -96,8 +97,8 @@ class Network:
         if self.get_client(client["client_id"]):
             return
 
-        logger.info("adding client {}".format(client["client_id"]))
-        self.client_store.upsert(client)
+        logger.info("adding client {}".format(client.to_dict()))
+        self.client_store.add(client)
 
     # TODO: This method is not used in the current version of FEDn.
     def get_client(self, client_id: str):
@@ -113,12 +114,3 @@ class Network:
             return client
         except Exception:
             return None
-
-    # TODO: This method is not used in the current version of FEDn.
-    def get_client_info(self):
-        """List available client in statestore.
-
-        :return: list of client objects
-        :rtype: list(ObjectId)
-        """
-        return self.client_store.list(limit=0, skip=0, sort_key=None)
--- a/fedn/network/api/v1/client_routes.py
+++ b/fedn/network/api/v1/client_routes.py
@@ -5,6 +5,7 @@ from fedn.common.log_config import logger
 from fedn.network.api.auth import jwt_auth_required
 from fedn.network.api.shared import client_store, control, get_checksum, package_store
 from fedn.network.api.v1.shared import api_version, get_post_data_to_kwargs, get_typed_list_headers
+from fedn.network.storage.statestore.models import Client
 
 bp = Blueprint("client", __name__, url_prefix=f"/api/{api_version}/clients")
 
@@ -354,9 +355,9 @@ def get_client(id: str):
                         type: string
     """
     try:
-        response = client_store.get(id)
+        response = client_store.get(id).to_dict()
         if response is None:
-          return jsonify({"message": f"Entity with id: {id} not found"}), 404
+            return jsonify({"message": f"Entity with id: {id} not found"}), 404
 
         return jsonify(response), 200
     except Exception as e:
@@ -399,7 +400,7 @@ def delete_client(id: str):
     try:
         result: bool = client_store.delete(id)
         if result is False:
-          return jsonify({"message": f"Entity with id: {id} not found"}), 404
+            return jsonify({"message": f"Entity with id: {id} not found"}), 404
 
         msg = "Client deleted" if result else "Client not deleted"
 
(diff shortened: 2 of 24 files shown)


fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -49,39 +42,82 @@ def get(self, id: str) -> Client:
...
+                del item_dict["id"]
+            result = self.database[self.collection].update_one({"client_id": item.client_id}, {"$set": item_dict}, upsert=True)
             id = result.upserted_id
             document = self.database[self.collection].find_one({"_id": id})
-            return True, from_document(document)
+            return True, Client(**from_document(document))
+        except Exception as e:
+            return False, str(e)
+
+    # TODO: This method has the same funcitonality as update but different signature. Should we change the signature of update and replace update with this?
```
niklastheman, line 93: Yes, the idea of the stores are they should all expose the same methods. Add, update etc...

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -91,7 +127,15 @@ def delete(self, id: str) -> bool:
 
         return super().delete(document["_id"])
 
-    def list(self, limit: int, skip: int, sort_key: str, sort_order=pymongo.DESCENDING, **kwargs) -> Dict[int, List[Client]]:
+    def select(self, **filter_kwargs) -> List[Client]:
```
niklastheman, line 130: Should be possible with list no? Would like to keep stores similar across all entities. They should all implement methods of the "base store" class... And IF anything extra is needed then expose additional methods. List not sufficient for all type of fetching? What's missing?

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -109,9 +153,15 @@ def list(self, limit: int, skip: int, sort_key: str, sort_order=pymongo.DESCENDI
...
+        """Count entities
+        param kwargs: Additional query parameters
+            type: dict
+            example: {"key": "models"}
+        return: The count of entities
+        """
         return super().count(**kwargs)
 
-    def connected_client_count(self, combiners: List[str]) -> List[Client]:
+    def connected_client_count(self, combiners: List[str]) -> List:
```
niklastheman, line 164: Why this change?

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -201,29 +246,109 @@ def update(self, id: str, item: Client) -> Tuple[bool, Any]:
...
+            if item is None:
+                return False
+
+            session.delete(item)
+            session.commit()
+
+            return True
+
+    # TODO: This method has the same funcitonality as update but different signature. Should we change the signature of update and replace update with this?
+    def commit(self, item: Client) -> Tuple[bool, Any]:
```
niklastheman, line 305: Same comment as above...

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -201,29 +246,109 @@ def update(self, id: str, item: Client) -> Tuple[bool, Any]:
...
+            existing_item.status = item.status
+            existing_item.last_seen = item.last_seen
+
+            session.commit()
+
+            return True, row_to_client(existing_item)
 
-    def delete(self, id):
-        raise NotImplementedError
+    def select(self, **filter_kwargs) -> List[Client]:
```
niklastheman, line 329: Same comment as above...

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -34,11 +22,16 @@ def upsert(self, item: Client) -> Tuple[bool, Any]:
     def connected_client_count(self, combiners: List[str]) -> List[Client]:
         pass
 
+    @abstractmethod
```
niklastheman, line 25: If upsert is no longer used, we can remove it. Commit => update it sounds like? No need for an extra method I believe...

fedn/network/storage/statestore/stores/client_store.py
carl-andersson: It does not return a list of clients, rather a list of dicts

niklastheman: Cool

carl-andersson:
Unified the framework more now, should be quite easy to migrate most of the other storage types now. I believe this still needs some testing, atleast the routes. 

Potential dangers when this goes into prod is that the mongo db contains more fields than I know of, i.e. residues from other features

fedn/network/api/v1/client_routes.py
```diff
@@ -486,17 +492,18 @@ def add_client():
...
-        control.network.add_client(client_config)
+        if client_store.get(client_id) is not None:
+            logger.info("adding client {}".format(client_id))
+            new_client = Client(
+                client_id=client_id,
+                name=name,
+                combiner=combiner.name,
+                combiner_preferred=preferred_combiner,
+                ip=remote_addr,
+                status="available",  # TODO: Should be "online"?
```
carl-andersson, line 503: Opinion here?

fedn/network/api/network.py
```diff
@@ -85,7 +86,8 @@ def handle_unavailable_combiner(self, combiner):
         # TODO: Implement strategy to handle an unavailable combiner.
         logger.warning("REDUCER CONTROL: Combiner {} unavailable.".format(combiner.name))
 
-    def add_client(self, client):
+    # TODO: This method is not used in the current version of FEDn.
```
carl-andersson, line 89: Should we clean up this?

carl-andersson:
New (unified) interface is in new_store.py, when all stores have migrated this should easily be changed back to store.py

niklastheman:
> Unified the framework more now, should be quite easy to migrate most of the other storage types now. I believe this still needs some testing, atleast the routes.
> 
> Potential dangers when this goes into prod is that the mongo db contains more fields than I know of, i.e. residues from other features

Perhaps an idea the would be to complete all stores prior to next release. Can we wait a week or two for next release @Wrede ?

carl-andersson:
I do think it would be a good idea to put it in a seperate branch. To simplify testing and reviewing. Is it possible to create pull request towards feature branches? 

I also think we should review this first change more closely as it will be used as foundation for the other rewrites.

fedn/network/api/network.py
niklastheman: Yes.

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -123,6 +26,44 @@ def connected_client_count(self, combiners: List[str]) -> List[Client]:
...
+
+    def get(self, client_id: str) -> Client:
+        entity = MongoDBStore.get(self, client_id)
+        if entity is None:
+            return None
+        return Client(**from_document(entity))
+
+    def add(self, item: Client) -> Tuple[bool, Any]:
+        item_dict = item.to_dict(exclude_unset=False)
+        success, obj = MongoDBStore.add(self, item_dict)
```
niklastheman, line 44: Why this implementation: MongoDBStore.add(self, item_dict)? MongoDBClientStore inherits from MongoDBStore and should therefore have access via super?

This an improvements somehow? If so, how :)?

fedn/network/storage/statestore/stores/client_store.py
```diff
@@ -140,166 +81,51 @@ def connected_client_count(self, combiners: List[str]) -> List[Client]:
...
-            return {"count": count, "result": result}
-
-    def count(self, **kwargs):
+    def delete(self, id) -> bool:
         with self.Session() as session:
-            stmt = select(func.count()).select_from(ClientModel)
-
-            for key, value in kwargs.items():
-                stmt = stmt.where(getattr(ClientModel, key) == value)
+            return SQLStore.delete(self, session, id)
```
niklastheman, line 117: Similar question as the one regarding the mongo store.

fedn/network/storage/statestore/stores/new_store.py
```diff
@@ -0,0 +1,224 @@
...
+        """Delete an entity.
+
+        param id: The id of the entity
+            type: str
+        return: A boolean indicating success or failure
+        """
+        pass
+
+    @abstractmethod
+    def select(self, limi: int = 0, skip: int = 0, sort_key: str = None, sort_order=pymongo.DESCENDING, **kwargs) -> List[T]:
```
niklastheman, line 62: Maybe call it list?

niklastheman (approved):
Sweet, these are good changes! I ran it locally and everything worked as expected. I have some questions.

fedn/network/storage/statestore/stores/new_store.py
carl-andersson: Sure, if we are content that it has changed functionality from before with the same name.

fedn/network/storage/statestore/stores/client_store.py
carl-andersson: I removed store inheritence for the base mongostore and sqlstore for two reasons,
1) the signature of the functions are not the same in these calls as the calls to the Store class and is unique for the base sql/mongo classes
2) the Store class is already inherited from the other class, i.e. ClientStore in this case. The main purpose of this inheritence as I see it is to provide doc and types 

With this removed inheritence you need to specify which super.get/update/list (the one from the abstract class or the one from the implementation)

fedn/network/storage/statestore/stores/new_store.py
niklastheman: I think this is a good improvement. Doesn't have to be changed now but this is what we want :)

fedn/network/storage/statestore/stores/client_store.py
niklastheman: Okey, cool. But then it sound like this should be reverted in a later stage no? Of course base class signature can be changed, but cool, let's wait a bit...

fedn/network/storage/statestore/stores/new_store.py
carl-andersson: I believe I agree, the naming of this method is easy to change later.

fedn/network/storage/statestore/stores/client_store.py
niklastheman: In that case I think we should separate implementation and interface no? Store inherits from class and the use "static" methods of class instead of the once in inherited, or am I getting it wrong? Just seems a bit messy if so..

PR #819, opened 2025-02-24
Feature/SK-1439 

stefanhellander:
## Description
 Set environment variable FEDN_CONNECT_API_SECURE=false to enable insecure calls to the API server.


Wrede (changes requested):
I would suggest you check the ENV in config.py instead and then import it to files where it's needed. 

https://github.com/scaleoutsystems/fedn/blob/master/fedn/common/config.py

Wrede (approved):

PR #820, opened 2025-02-24
Bugfix/SK-1440 | Syntax error escape character on windows paths

Wrede:
Escape chars (back slash) raised error because os.path.join was not used in os.walk to find fedn.yaml, further trying to delete the pacjage tar while it was in with clause. This PR fixes that.


sztoor (approved):
Windows client is working with these changes.

PR #821, opened 2025-02-25
Feature/SK-1359 | Unit test get (single) methods for client

carl-andersson:


carl-andersson:
Any functionality that is missing a test?

fedn/tests/stores/test_client_store.py
```diff
@@ -61,16 +72,60 @@ def options():
 
 class TestClientStore:
 
-    def test_add_update_delete(self, postgres_connection:DatabaseConnection, sql_connection: DatabaseConnection, mongo_connection:DatabaseConnection):
-        pass
-
-    def test_list(self, db_connections_with_data: list[tuple[str, DatabaseConnection]], options: list[tuple]):   
+    def test_add_get_update_delete(self, postgres_connection:DatabaseConnection, sql_connection: DatabaseConnection, mongo_connection:DatabaseConnection):
```
niklastheman, line 75: Just a comment, but could be worth separating... Mostly just in case something does go wrong... But, this is big improvement regardless.

niklastheman (approved):
Cool, FEDn is getting better and better!

fedn/tests/stores/test_client_store.py
niklastheman: That and if you want dived or add others for add and update... BUT, this is just a comment... My thinking is that when a test fails, the less it is testing the easier it will be to solve... Again, this is a good improvement as is!

PR #822, opened 2025-02-25
Feature/SK-1435 | Update modelstore to use DTOs

carl-andersson:
Tests and conversion to DTO for models

Includes some changes to ClientStore also to streamline the implementation

### DIFF
--- a/fedn/network/api/v1/graphql/schema.py
+++ b/fedn/network/api/v1/graphql/schema.py
@@ -84,8 +84,8 @@ class SessionType(graphene.ObjectType):
 
     def resolve_models(self, info):
         kwargs = {"session_id": self["session_id"]}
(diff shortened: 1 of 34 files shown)


fedn/network/api/v1/graphql/schema.py
niklastheman, line 1: Tested graphql after these updates?

fedn/network/api/v1/session_routes.py
```diff
@@ -393,6 +393,7 @@ def start_session():
         if nr_available_clients < min_clients:
             return jsonify({"message": f"Number of available clients is lower than the required minimum of {min_clients}"}), 400
 
+        # TODO: This line is a noop now since get returns None if the model is not found
```
niklastheman, line 396: Good catch can we fix this in this PR?

niklastheman (approved):

PR #823, opened 2025-02-26
Bugfix/SK-1452 | GrpcConnectionOptions all args not needed for direct combiner connection

Wrede:
GRPCConnectionOptions returned missing constructor arguments error when connecting directly to combiner without service discovery at controller, i.e --combiner and --combiner-port in CLI.


benjaminastrand (approved):
LGTM!

PR #824, opened 2025-02-26
Fix/SK-1451 | Check if there is an active helper before using "numpyhelper" as default

benjaminastrand:
When starting session in APIClient, the helper now comes from the active package if the user hasn't specified a helper when invoking start_session(). Otherwise it uses numpyhelper as default.

### DIFF
--- a/fedn/network/api/client.py
+++ b/fedn/network/api/client.py
@@ -583,7 +583,7 @@ class APIClient:
         round_buffer_size: int = -1,
         delete_models: bool = True,
         validate: bool = True,
-        helper: str = "numpyhelper",
+        helper: str = None,
         min_clients: int = 1,
         requested_clients: int = 8,
         server_functions: ServerFunctionsBase = None,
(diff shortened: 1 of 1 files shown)


niklastheman (approved):

fedn/network/api/v1/helper_routes.py
```diff
@@ -27,7 +27,7 @@ def get_active_helper():
     try:
         active_package = package_store.get_active()
         if active_package is None:
-            return jsonify({"message": "No active helper"}), 404
+            return jsonify({"message": "No active helper"}), 400
```
niklastheman, line 30: I think 404 is preferred. It's not a bad request no? Helper is simply not set...

niklastheman (changes requested):

fedn/network/api/v1/helper_routes.py
benjaminastrand: Valid point, switching back

niklastheman (approved):

PR #832, opened 2025-03-03
Feature/SK-1462 | Add Combiner DTO

carl-andersson:
* Converted Combiner store to use DTO for transactions to the database
* Added tests for Combinerstore

### DIFF
--- a/fedn/network/storage/statestore/stores/combiner_store.py
+++ b/fedn/network/storage/statestore/stores/combiner_store.py
@@ -1,5 +1,5 @@
 from abc import abstractmethod
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List
 
 import pymongo
 from pymongo.database import Database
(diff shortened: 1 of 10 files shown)


fedn/network/storage/statestore/stores/combiner_store.py
```diff
@@ -1,217 +1,107 @@
...
-        "fqdn": row.fqdn,
-        "port": row.port,
-        "updated_at": row.updated_at,
-    }
+    def _dto_from_document(self, document: Dict) -> CombinerDTO:
+        return CombinerDTO().populate_with(from_document(document))
 
 
-class SQLCombinerStore(CombinerStore, SQLStore[Combiner]):
+class SQLCombinerStore(CombinerStore, SQLStore[CombinerDTO]):
```
niklastheman, line 61: Missing the get_by_name method? Is this solved in another way? If so great, we should remove get_by_name from the interface...

niklastheman (approved):
Great! I have one question but approved anyhow...

fedn/network/storage/statestore/stores/combiner_store.py
carl-andersson: It should be there? It is a new method, so added it to both mongo and sql. I believe it is required but could be discussed.

niklastheman: Great! Yes, I think in the future we can think of removing it... Name should not be used as key I think...

PR #837, opened 2025-03-05
Feature/SK-1473 | Store memory and cpu utilisation of clients

niklastheman:
Send cpu and memory utilisation via send heart beat


ahellander (approved):
Looks good. Maybe it should be named something else that "Analytics" in the future, but works for now.

Wrede (approved):
Approved but I don't like the name Analytic

PR #840, opened 2025-03-06
Feature/SK-1479 | Limit analytics db size

niklastheman:
Remove analytics records older than 5 min


carl-andersson (commented):
Looks good

carl-andersson (approved):

PR #846, opened 2025-03-18
Bug/SK-1495 | Start session from api client has incorrect documentation

niklastheman:
### Api Client docs

start_session id => name


Hovstadius (approved):
Tested and it looks good!

PR #849, opened 2025-03-24
Feature/SK-1263 | Improve the API-Client example

niklastheman:
# Mnist-Pytorch local package example

Notebook added that guides users in converting their standard mnist-pytorch model into a Federated setting.


ahellander:
Nice addition with the new local package notebook. If we refine it a bit, we could put it on colab and then ppl could try a fedn client really easily.

niklastheman:
> Nice addition with the new local package notebook. If we refine it a bit, we could put it on colab and then ppl could try a fedn client really easily.

Yes sounds good!

niklastheman:
> # Mnist-Pytorch local compute example
> Notebook added that guides users in converting their standard mnist-pytorch model into a Federated setting.

mattiasakesson (approved):
Now it looks refined and good to merge

PR #855, opened 2025-03-27
Feature/SK-1496 | Integrated validation feedback via exceptions

carl-andersson:
* Generalized many calls to the database to reduce the amount of boiler plate
  * This includes new two new nessesary functions for stores, to convert to and from dto to store objects. 
  * This are called _dto_from_document and _document_from_dto and similarly for SQL
  * Avoided using add: super().add() which is arguable
* Added PrimaryID field to dto to be able to flag a field as a primary key and access it generically
* Implmented validations for DTOs via a decorator called validator
* Added ValidationError and MissingValueError
* Added catches to add and update API calls to capture the new exceptions and give user friendly feedback


carl-andersson:
I believe I've quenched the warning messages from copilot about leaking exceptions by defining a user message.

niklastheman (approved):
Awesome!

PR #857, opened 2025-03-27
Docs/SK-1506 | API References update

carl-andersson:
* New Makefile
* Checked ClientApi getting started for inconsistencies


Wrede (changes requested):
Protobuf generated files seems to be in api reference

Wrede (approved):

PR #865, opened 2025-04-02
Feature/SK-1523 | Add endpoint to get models with on children

niklastheman:
- added get_leaf_nodes endpoint to /models
- bug fix for ancestors
- descendants endpoints
- get_leaf_nodes added to model store


carl-andersson (approved):
Looks good

PR #867, opened 2025-04-03
Docs/SK-1509 | Documentation for Local Compute 

FrankJonasmoelle:
Adds Documentation for Local Compute notebook example, which allows a user to run FEDn clients directly in scripts.


niklastheman (approved):
Awesome, I like it! I like it a lot!

PR #868, opened 2025-04-04
Feature/SK-1519 | FEDn container uses distroless image

stefanhellander:
## Description

Switch to a "distroless" base image.

Note that it will not be possible to run shell scripts in this image, so integration tests are not working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
-ARG BASE_IMG=python:3.11-slim
+# ARG BASE_IMG=python:3.11-slim
 ARG RUNTIME_IMG=gcr.io/distroless/python3
-FROM $BASE_IMG AS builder
+# FROM $BASE_IMG AS builder
+FROM python:3.11-slim as builder
 
 ARG GRPC_HEALTH_PROBE_VERSION=""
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,77 +1,24 @@
-# Stage 1: Builder
-ARG BASE_IMG=python:3.12-slim
-FROM $BASE_IMG AS builder
+FROM python:3.11-slim AS builder
```
Wrede, line 1: why downgrade python?

Dockerfile
```diff
@@ -1,77 +1,24 @@
...
-  fi \
-  && rm -rf /build/requirements.txt
-
 
-# Install grpc health probe
-RUN if [ ! -z "$GRPC_HEALTH_PROBE_VERSION" ]; then \
-  wget -qO /build/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
-  chmod +x /build/grpc_health_probe; \
-  fi
+RUN mkdir /python-dist \
```
Wrede, line 12: where is fedn installed?

Wrede (commented):
On your note, there are no shell scripts running in the image? In the CI tests scripts are run in ubuntu images provided by github.

Dockerfile
stefanhellander: RUN mkdir /python-dist \
  && pip install --upgrade pip \
  && pip install --prefix=/python-dist --no-cache-dir 'setuptools>=65' .

PR #869, opened 2025-04-04
Bug/SK-1527 | Active compute package not set

niklastheman:
Set active property in list method


Wrede:
Fixed in release v0.25.2

PR #870, opened 2025-04-04
Bug/SK-1529 | save_metric fails on cli client

carl-andersson:
* Revert change that was supposed to fix unicode issues


Wrede (approved):

Wrede:
Fixed in release v0.25.2

PR #877, opened 2025-04-09
Bug/SK-1537 | Unable to stop session

niklastheman:
Session patch endpoint updated

### DIFF
--- a/fedn/network/api/v1/session_routes.py
+++ b/fedn/network/api/v1/session_routes.py
@@ -485,6 +485,7 @@ def patch_session(id: str):
         data = request.get_json()
         data["session_id"] = id
 
+        session.patch_with(existing_session.to_dict(), throw_on_extra_keys=False)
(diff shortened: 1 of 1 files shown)


fedn/network/api/v1/session_routes.py
```diff
@@ -485,8 +485,8 @@ def patch_session(id: str):
         data = request.get_json()
         data["session_id"] = id
 
-        session.patch_with(data, throw_on_extra_keys=False)
-
+        merged_data = {**existing_session.to_dict(), **data}
+        session.patch_with(merged_data, throw_on_extra_keys=False)
```
carl-andersson, line 489: I believe a neither solution would be to patch the existing session 
```suggestion
        existing_session.patch_with(data, throw_on_extra_keys=False)
```

carl-andersson (approved):

PR #878, opened 2025-04-09
Feature/SK-1541 | Add update project cli feature

KatHellg:


Wrede (approved):
nice! Obs that you have many PR on similar code locations. Depending on merge order you will get conflict. Consider merging into a single feature branch instead (collect all merges into one branch and then create a collected PR into main).

PR #879, opened 2025-04-09
Bug/SK-1544 | Models PATCH updates incorrectly

niklastheman:
- patch endpoint fix
- update models key only


carl-andersson (approved):
Looks, good. Consider Copilots comment though

PR #883, opened 2025-04-14
Fix/SK-1547 | Fixing Huggingface train.py imports

FrankJonasmoelle:
There is an issue where AdamW is imported from the Transformers library, which is a mistake. 

The changes remove this and use AdamW directly from torch (torch.optim.AdamW)


carl-andersson (approved):
👍

PR #900, opened 2025-04-24
Feature/SK-1578 | Add tests for server functions

viktorvaladi:
Adds server functions to integration-test

Adds server functions to integration-test-studio

Tests work by adding the server functions example to the PYTHONPATH, so that the api tests can import it and send it to start functions.


niklastheman (approved):
Great to cover this in testing!

Wrede (approved):
Need careful monitoring after merge

PR #910, opened 2025-05-05
Feature/SK-1579 | Add a field 'updated_at' similar to 'committed_at'

niklastheman:
updated_at field added to all entities

### DIFF
--- a/fedn/network/storage/statestore/stores/sql/shared.py
+++ b/fedn/network/storage/statestore/stores/sql/shared.py
@@ -30,6 +30,7 @@ class MyAbstractBase(Base):
 
     id: Mapped[str] = mapped_column(primary_key=True, default=lambda: str(uuid.uuid4()))
     committed_at: Mapped[datetime] = mapped_column(default=datetime.now)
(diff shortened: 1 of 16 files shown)


carl-andersson (approved):
sql_update should also update the updated_at field

fedn/network/storage/statestore/stores/sql/shared.py
```diff
@@ -30,6 +30,7 @@ class MyAbstractBase(Base):
 
     id: Mapped[str] = mapped_column(primary_key=True, default=lambda: str(uuid.uuid4()))
     committed_at: Mapped[datetime] = mapped_column(default=datetime.now)
+    updated_at: Mapped[datetime] = mapped_column(default=datetime.now, onupdate=datetime.now)
```
niklastheman, line 33: @carl-andersson this should take care of the sql updates no?

niklastheman:
> * sql_update should also update the updated_at field
> * Most tests need to ignore update_at field

see comment above

fedn/network/storage/statestore/stores/sql/shared.py
carl-andersson: Yeah, sorry missed that

PR #912, opened 2025-05-05
Feature/SK-1595 | Api client for server functions + example

viktorvaladi:
Adds initialized api client to sf through allowed imports. Through this users can use any APIClient functionality.

Example show getting client attributes (beta functionality) and using them for client selection.

Example uses beta version of sending client attribute through the APIClient.


examples/server-functions/send_attributes.py
```diff
@@ -0,0 +1,37 @@
...
+
+def main():
+    # --------- NOTE ------------
+    # run this script with --token <admin token> (fetch admin token from studio)
+    # --api-url
+    args = parse_args()
+    client = APIClient(host=args.api_url, token=args.admin_token, secure=True, verify=True)
+    while True:
+        # Prepare a random charging status
+        attribute_payload = {"key": "charging", "value": random.choice([True, False]), "sender": {"name": "", "role": "", "client_id": args.client_id}}
```
carl-andersson, line 25: Maybe have the choice to "True", "False" since the attribute value is a string

examples/server-functions/server_functions.py
```diff
@@ -13,10 +13,29 @@ def __init__(self) -> None:
...
+            print(f"Warning: unable to fetch attributes ({e}), selecting all clients")
+            return client_ids
+
+        selected = []
+        charging_count = 0
+        for cid in client_ids:
+            client_attrs = attrs_map.get(cid) or {}
+            charging = client_attrs.get("charging", None)
+            # Default to select to not depend on beta version of attributes.
+            if charging is True or charging is None:
```
carl-andersson, line 32: Charging will never be True? For the same reason as above, it will always be string

fedn/network/api/client.py
```diff
@@ -975,3 +975,20 @@ def get_current_attributes(self, client_list):
...
         _json = response.json()
         return _json
+
+    def add_attributes(self, attribute: dict) -> dict:
+        """Add or update client attributes via the controller API.
+
+        :param attribute: A dict matching AttributeDTO.schema, e.g.
+            {
+                "key": "charging",
+                "value": true,
```
carl-andersson, line 985: Here also

carl-andersson (approved):
I think that you need to address the problem with types here since attributes are strings

examples/server-functions/send_attributes.py
viktorvaladi: It is comming out as a bool when fetching it so this seems to work.

examples/server-functions/server_functions.py
viktorvaladi: Same as above, it is comming out as True or False (bools) :)

fedn/network/api/client.py
viktorvaladi: same :D

examples/server-functions/server_functions.py
```diff
@@ -13,10 +13,29 @@ def __init__(self) -> None:
...
+                selected.append(cid)
+                charging_count += 1
+
+        print(f"Selected clients: {selected}")
+        print(f"{charging_count} clients selected based on client attributes, out of {len(client_ids)} clients.")
```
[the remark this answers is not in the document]
viktorvaladi: since this is user defined code we dont want the user to have to know to use a logger

examples/server-functions/send_attributes.py
carl-andersson: I understand, we should change that in that case. It is not intended and will not work the same if attributes are added with e.g. GRPC or using an other DB

examples/server-functions/send_attributes.py
```diff
@@ -0,0 +1,37 @@
...
+
+        # Send to server
+        try:
+            result = client.add_attributes(attribute_payload)
+            print(json.dumps(result, indent=2))
+        except Exception as e:
+            print(f"Failed to send attributes: {e}")
+        time.sleep(30)
```
Wrede, line 33: I would add the sleep as a parser arg

fedn/network/combiner/hooks/hooks.py
```diff
@@ -50,7 +52,7 @@ def HandleClientConfig(self, request_iterator: fedn.ClientConfigRequest, context
             logger.info(f"Client config response: {client_settings}")
             return fedn.ClientConfigResponse(client_settings=json.dumps(client_settings))
         except Exception as e:
-            logger.exception("Error handling client config request: %s", e)
+            logger.exception(f"Error handling client config request: {e}")
```
Wrede, line 55: use logger.error

examples/server-functions/send_attributes.py
viktorvaladi: Ok, updated.

Wrede (approved):
approved with comments

Wrede (approved):
approved with comments

PR #918, opened 2025-05-08
Bugfix/SK-1602 | Telemetries are not getting cleared

carl-andersson:
* Corrected a bug were telemetries are not getting deleted after 5 min
* Fixed a bug if heartbeats were renabled would send them too fast

### DIFF
--- a/fedn/network/clients/grpc_handler.py
+++ b/fedn/network/clients/grpc_handler.py
@@ -131,6 +131,7 @@ class GrpcHandler:
         while send_heartbeat:
             try:
                 response = self.heartbeat(client_name, client_id)
+                time.sleep(update_frequency)
             except grpc.RpcError as e:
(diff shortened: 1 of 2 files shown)


fedn/network/clients/grpc_handler.py
```diff
@@ -131,6 +131,7 @@ def send_heartbeats(self, client_name: str, client_id: str, update_frequency: fl
         while send_heartbeat:
             try:
                 response = self.heartbeat(client_name, client_id)
+                time.sleep(update_frequency)
```
niklastheman, line 134: This is unrelated to the bug in title?

niklastheman (approved):

fedn/network/clients/grpc_handler.py
carl-andersson: Yes, but I think it's good to get it out

niklastheman: Sweet, yes let's!

PR #991, opened 2025-09-12
Feature/SK-1659 | Enable combiner to control tasks on client

carl-andersson:
* Combiner now notifies the client if a task has been cancelled or timedout
  * Requires the polling client
  * Upto the user to call check_task_abort regularly
* Reworked the client state store on the combiner to a ClientInterface to improve code quality


Wrede (approved):
Really like the clientManager and ClientInterface it becomes much more clearer to understand.

PR #996, opened 2025-09-17
Fix/SK-1727 | Change to urljoin

viktorvaladi:
Change so that api url for client connection uses urljoin, making it more robust to different string formats.


carl-andersson (approved):
😃

PR #1021, opened 2026-03-30
Update links

blaudden:
Remove readthedocs build image
Remove github workflow image
Use correct docs link


blaudden:
Abandoning this and making new so we see no checks run

